PR #6, opened 2024-08-27
WIP 

stefnotch:
I opened the PR so that I can leave comments right here on GitHub


README.md
```diff
@@ -1,480 +1,133 @@
...
+[WGSL](https://www.w3.org/TR/WGSL/) shading language
+using community tooling.
+We want to see WGSL enhancements shared more broadly.
+We want to enable a WGSL library ecosystem.
 
-We propose adding an importing mechanism to the WGSL shading language as an extension.
+Many community projects have independently uncovered
+a need for WGSL enhancements like module composition,
+conditional compilation,
+and runtime variable substitution.
```
stefnotch, line 12: What do you mean with "runtime variable substitution"?

README.md
```diff
@@ -1,480 +1,133 @@
...
+A WGSL enhancement for `import` statements
+to enable importing WGSL elements like functions from other files is our first priority.
+See spec draft here: **[imports](./Imports.md)**.
 
-Finally, we want **multiple tools** which can compile WGSL-with-imports down to raw WGSL. Using WGSL-with-imports both in Rust projects and in web projects should be possible.
+Several other WGSL enhancements
+are on the [roadmap](#wgsl-enhancements-roadmap).
 
-## Guide-level explanation
+## Tools
```
stefnotch, line 47: This section is amazing. 👍

Imports.md
```diff
@@ -0,0 +1,407 @@
...
+
+An import statement is parsed as follows, with spaces and comments allowed between tokens:
+
+```
+main:
+| 'import' import_path ';'
+
+import_path:
+| ('.' | '..') '/' import_path
+| ident '/' (import_path | import_collection | item_import | star_import)
```
stefnotch, line 77: I think we could restrict the dots by writing

```
main:
| 'import' import_relative? import_path ';'

import_relative:
| ('.' | '..') '/' ('..' '/')*

import_path:
| ident '/' (import_path | import_collection | item_import | star_import)

```

NameMangling.md
```diff
@@ -0,0 +1,37 @@
+# Name mangling
+
+Linker will mangle names internally,
+but we're hoping to keep the mangling details
+out of the public api.
+
+See also renaming host visible names in [Visiblity](./Visiblity.md).
```
stefnotch, line 7: Typo, it should be `[Visibility](./Visibility.md)`

NameMangling.md
```diff
@@ -0,0 +1,37 @@
...
+
+(TBD)
+
+## Mangling
```
stefnotch, line 11: Interestingly, this standard WGSL mangling scheme might need to be extended to support
- WGSL modules that are outside of the package root (not sure if they should be supported)
- Generics 😎

VariableSubstiution.md
```diff
@@ -0,0 +1,27 @@
+# Variable Substitution
```
stefnotch, line 1: I think the name of this file has a typo

stefnotch (commented):
This is amazing, thank you very much! @mighdoll 
I pretty much agree with everything here, and think it's really helpful to have this written down. There are some sections where I have some initial thoughts, but I'll have to think more about topics other than importing 👍

ConditionalCompilation.md
```diff
@@ -0,0 +1,15 @@
+# Conditional Compilation
+
+(TBD)
+
+This section will discuss extensions to WGSL
+to support conditional compilation, where the conditions
+are set by the host code runtime or build time.
+
+* `#if #else #endif` (or perhaps `#ifdef`) in the style of a C preprocessor
```
ncthbrt, line 9: I'd really prefer the rust style as it's so much more structured. It would make optimizations a lot easier. For example could parse/lower certain functions ahead of time and easily remove them from the IR vs `#if #else` which operates on a source level.

Extends.md
```diff
@@ -0,0 +1,20 @@
+# Struct Inheritance for WGSL
+
+(TBD)
+
+It’s convenient to be able to express structures as combinations of other structures as in Java/TypeScript.
+
+wgsl-linker currently supports an `extends` syntax for struct inheritance (but `extends` is only lightly
```
ncthbrt, line 7: Interesting. Extends kind of has the same semantics here but on a struct level as what I was proposing for modules

Extends.md
```diff
@@ -0,0 +1,20 @@
...
+by referencing through a member field. But:
+
+* Using fields of the referenced struct requires an additional field dereference "." in the syntax
+  for each level of indirection.
+* Some abstractions may be hard to express w/o inheritance (since we also don't have interfaces in wgsl).
+
+Issues:
+
+* TBD demonstrate an example or drop the feature.
+* Also, would generics for structs be an alternate way to parameterize composition?
```
ncthbrt, line 20: I like the idea of struct/module "shapes". They are interfaces in a sense, but don't bring in OO ideas and implementation of shapes could be structural rather than explicit. Module shapes in particular would be kind of like rust traits, but instead of operating on a struct, they'd operate on a module, which would become the unit of composition in this extended wgsl. I also like the idea of generic functions and structs though am nervous about what they'd do to compile times based on the inference technique chosen.

Generics.md
```diff
@@ -0,0 +1,164 @@
...
+        return acc;
+      }
+    }
+
+    fn op_add<E>(e1: E, e2: E) {
+      return e1 + e2;
+    }
+
+    fn array_sum<E, N>(elems: array<E, N>) -> E {
+      return accumulate<E, op_add<E>, N>(0, elems);
```
ncthbrt, line 158: Could you not get around that by doing a cast using to `E()`?

Imports.md
```diff
@@ -0,0 +1,407 @@
...
+By placing an import at the very top of a file, one can either import an entire module, or only specific importable items, such as functions, structs or types.
+
+```
+// Importing a single item using a relative path
+import ./lighting/pbr;
+
+// Importing multiple items
+import my/geom/sphere/{ draw, default_radius as foobar };
+
+// Imports a whole module. Use it with `bevy_ui.name`
```
ncthbrt, line 33: This seems semantically a bit odd. Wouldn't it make more sense to have this syntax just bring in name and the other symbols into the current scope?

Imports.md
```diff
@@ -0,0 +1,407 @@
...
+The `import` statement extension is designed to appear somewhat familiar to TypeScript syntax,
+but with a Rust like recursive grammar to make importing
+multiple importable items less verbose.
+The syntax is also heavily inspired by [Gleam](https://gleam.run/).
+
+By placing an import at the very top of a file, one can either import an entire module, or only specific importable items, such as functions, structs or types.
+
+```
+// Importing a single item using a relative path
+import ./lighting/pbr;
```
ncthbrt, line 28: In contrast, this feels like I'm importing a module and could call its members using `pbr.<module_member_name>`. 

Also using a dot to refer to an item inside a module seems like it might introduce ambiguity and additional challenges with the rest of wgsl?

Imports.md
```diff
@@ -0,0 +1,407 @@
...
+
+```
+import bevy_pbr/{
+  forward_io/VertexOutput,
+  pbr_types/{PbrInput, pbr_input_new},
+  pbr_bindings/*
+};
+
+fn main() {
+
```
ncthbrt, line 61: missing an example of usage here?

Imports.md
```diff
@@ -0,0 +1,407 @@
...
+}
+```
+
+# Reference-level explanation
+
+Imports must appear as the first items in a WGSL file. They import "importable items" (see [GLOSSARY.md](./GLOSSARY.md)).
```
ncthbrt, line 67: We might want to later introduce the idea of scoped imports?

Imports.md
```diff
@@ -0,0 +1,407 @@
...
+- [Global diagnostic filters](https://www.w3.org/TR/WGSL/#global-diagnostic-directive)
+
+These are always set at the very top in the main module, and affect all imported modules. They come before the imports.
+
+When, during parsing of imported modules, we encounter an extension or a global diagnostic filter, we check if the main module enables it, or sets the filter.
+If yes, everything is fine. If not, we throw an error.
+
+## Preserved items
+
+These items are preserved when importing a module. They are not imported, but will land in the final module with a mangled name.
```
ncthbrt, line 225: In naga-oil entrypoints are lowered to normal functions. I'm not sure if that's the correct semantics and is against the spec, but just wanted to comment on current behaviour that is being used in the wild.

NameMangling.md
```diff
@@ -0,0 +1,37 @@
...
+To join multiple modules into one, we need to make sure that names do not collide. This is done by name mangling.
+
+Compiled WGSL files have a few exposed parts, namely the names of the entry functions and pipeline overridable constants. These need to be stable and predictable, so we will introduce a stable, predictable and human-readable name mangling scheme.
+
+Finally, only importable items need to be mangled.
+
+For name mangling, we introduce the concept of a **absolute module identifier**. This is an ID that is relative to the nearest _package_, and is unique for each module. For example, given a file `my/geom/sphere.wgsl`, and a package `my`, then the absolute module identifier is `my_geom_sphere`. It is always this, regardless of how the file is imported. If a file or folder name already contains an underscore, it is replaced by two underscores.
```
ncthbrt, line 18: Much nicer than naga-oil!

VariableSubstiution.md
```diff
@@ -0,0 +1,27 @@
...
+Issues
+
+* Default values should be visible in the WGSL, so that language servers have something to go on
+* `#define`?
+* special syntax for names to be replaced, e.g. `#{WORKGROUP_SIZE}` or just an all caps convention like C?
+* Process as a preprocessor step, before other WESL processing.
+* Runtime modifiable variables are visible to the host, and so changes to names affect the api of the wgsl package.
+  Can we make the list of names available so that host code can autocomplete and validate names?
+* Are values opaque strings
+* should we somehow be integrating or extending override?
```
ncthbrt, line 27: I like the idea of extending override. But currently don't have the imagination to determine how exactly this should look

Versioning.md
```diff
@@ -0,0 +1,13 @@
...
+(TBD)
+
+As we stabilize for 1.0, we should mark a way for tools (and users) to distinguish our first stable version
+from subsequent releases.
+
+Issues:
+
+* Feature flags or versions?
+* perhaps `version WESL2024`?
+* Does the version need to go in wgsl or just in metadata like wgsl.toml.
```
ncthbrt, line 12: Would prefer for libraries that not every file needs to contain this. Seems an annoyance

Visibility.md
```diff
@@ -0,0 +1,84 @@
...
+
+```
+import my/lighting/{ pbr };
+
+export { pbr };
+```
+
+### Exporting as an attribute
+
+Exports can also be added as an attribute to the item itself.
```
ncthbrt, line 49: Why not `export struct Foo`. Many current wgsl parsers would choke on the unknown attribute as is and feels like having two export forms is a bit inconsistent.

Imports.md
stefnotch: It pretty much just mirrors the JavaScript star imports. 
`import * as bar from "foo/bar.wgsl";`
Would directly be translated to
`import foo/bar* as bar;`
Since that uselessly repeats `bar`, the shorthand is `import foo/bar/*;`

As for why
- We couldn't do `import foo/bar` for importing modules, because of the ambiguity issues.
- Rust-style star imports require shadowing.
- So we copied whatever Typescript does, plus a shorthand.

Imports.md
stefnotch: I looked through the parsing and implemented it. It's unambiguous. 🎉 

I'm happy with requiring curly braces for importing items though. Maybe that actually makes them easier to read.

Imports.md
stefnotch: Seems like a reasonable idea :)

Versioning.md
stefnotch: Yep, I think we'd all agree on that. Having a custom file extension + a version declaration in the main file/the wgsl.toml file should be good. I do really like the Rust concept of editions, so I personally would like to copy that.

README.md
mighdoll: I meant replacing wgsl text `#{BLOCK_SIZE}` with a value set at runtime, like `16`.

Need a better name tho, clearly. hmm.. maybe 'simple templating'?

ConditionalCompilation.md
mighdoll: yep, adding to discussion session. (also I agree that the c processor style is clunkier and we should consider rust style alternative)

Imports.md
ncthbrt: Could we not disambiguate item imports by making them have to be surrounded by braces?

VariableSubstiution.md
mighdoll: renamed file to SimpleTemplating

NameMangling.md
mighdoll: sharp eyed catch

NameMangling.md
mighdoll: Adding comment re: generics.

agreed, re limiting imports to within the package_root. If users want to import from an uncle dir, they can move the package_root up one dir.

README.md
```diff
@@ -1,480 +1,133 @@
...
+e.g. `#if MOBILE_CLASS_GPU`
+or `#[cfg gpu_class="mobile")]`.
 
-[dependencies]
-bevy_ui = { cargo = "bevy_ui" }
-shader_wiz = { path = "./node_modules/shader_wiz/src/main.wgsl" }
-```
+* **[simple templating](./SimpleTemplating.md)** -
+Insert runtime
+variables into WGSL from Rust/JavaScript host code into WGSL. e.g. `#{WORKGROUP_SIZE}`,
```
ncthbrt, line 98: I think these should be set per package and able to be defined in the manifest file. Otherwise you're leaking namespaces in your dependency tree that might result in unintended side effects

NameMangling.md
mighdoll: tho I'm still hoping we can hide mangling from the api, leaving it as an implementation detail.
will add comment

README.md
mighdoll: yea!

Extends.md
mighdoll: looking forward to understanding the modules idea, more to read.

Extends.md
mighdoll: yah, I'm hoping that we can do a simple form of generics w/o inference, to make things easier.

Imports.md
stefnotch: We could ! I actually tried it out, and was unsure about it.
So one of the examples (which is based on code in one of the Bevy shaders) would look like this
```
import bevy_pbr/{
  forward_io/{VertexOutput},
  pbr_types/{PbrInput, pbr_input_new}
};
import bevy_pbr/pbr_bindings; // needed to be moved out of the statement above
```

ncthbrt: I like that more. Would also bring back the possibility of importing all symbols into the current module without having to do `module_alias.item`

stefnotch: Actually, back whe I tried it out, I was using a more Rust-like syntax. So
```
import bevy_pbr::{
  forward_io::{VertexOutput},
  pbr_types::{PbrInput, pbr_input_new}
};
import bevy_pbr::pbr_bindings;
```

That kept tripping me up, because I kept thinking "ah, Rust style imports". So I always forgot that the curly braces are not optional.

At least in this variation, there's no risk of mistaking the syntax for Rust.

ncthbrt:
@stefnotch @mighdoll: what if we consider unifying semantics rather than syntax "flavour" and have a common manifest and built library output that allows seamless interoperability between the flavours? Could share a lot of code that way!

Like you could for example have a ts flavour and a rust flavour and both would just "work". 

For example, we could have a symbol table with visibility and the ts flavour could read that and visa versa.

Generics.md
k2d222: aah yes, that makes sense :)

README.md
```diff
@@ -1,480 +1,133 @@
...
 
-// Importing multiple items
-import my/geom/sphere/{ draw, default_radius as foobar };
+**WGSL Language Servers** - Enable syntax directed editing in IDEs.
+Shader development with IDEs should be well supported.
 
-// Imports a whole module. Use it with `bevy_ui.name`
-import bevy_ui/*;
-```
+**Syntax Highlighters** - Display pretty extended WGSL on web pages with
```
k2d222, line 59: People will also be interested in a tree-sitter grammar and/or a kate grammar for IDEs like vscode.

Generics.md
k2d222: yup, good find :)

Imports.md
mighdoll: Requiring curly braces at the end seems nice for people learning the syntax. 
perhaps start with that (and could still relax the syntax in the future.)

mighdoll: adding this issue to TBD section

Imports.md
mighdoll: adding tbd entry!

Imports.md
mighdoll: added to future section

Imports.md
mighdoll: you're a grammar whiz. adding

Imports.md
mighdoll: will add a note

Imports.md
mighdoll: adding TBD item

README.md
mighdoll: changed to simple templating.

README.md
mighdoll: agreed. added comment on packaging page.

README.md
mighdoll: yah, good one. adding item

Versioning.md
mighdoll: will mention editions and drop question about wgsl

Visibility.md
mighdoll: added TBD item.

VariableSubstiution.md
mighdoll: i like the idea in principle, hoping inspiration will strike

mighdoll:
> @stefnotch @mighdoll: what if we consider unifying semantics rather than syntax "flavour" and have a common manifest and built library output that allows seamless interoperability between the flavours? Could share a lot of code that way!
> 
> Like you could for example have a ts flavour and a rust flavour and both would just "work".
> 
> For example, we could have a symbol table with visibility and the ts flavour could read that and visa versa.
> 
> Moving away from a rusty flavour I think would be a hard sell for bevy as I understand it.

Yeah, I'm sympathetic to this having heard consistent feedback to things look more like typescript. I suppose many js/ts folks are afraid of rust and so :: scares them. 

I even went a fair ways towards implementing dual rust/ts import syntaxes. But saner people have pointed out that it's a little crazy to take on that maintenance burden forever if we can possibly avoid it.

Generics.md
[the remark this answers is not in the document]
mighdoll: more to discuss on generics for sure.

ncthbrt:
> > @stefnotch @mighdoll: what if we consider unifying semantics rather than syntax "flavour" and have a common manifest and built library output that allows seamless interoperability between the flavours? Could share a lot of code that way!
> 
> > 
> 
> > Like you could for example have a ts flavour and a rust flavour and both would just "work".
> 
> > 
> 
> > For example, we could have a symbol table with visibility and the ts flavour could read that and visa versa.
> 
> > 
> 
> > Moving away from a rusty flavour I think would be a hard sell for bevy as I understand it.
> 
> 
> 
> Yeah, I'm sympathetic to this having heard consistent feedback to things look more like typescript. I suppose many js/ts folks are afraid of rust and so :: scares them. 
> 
> 
> 
> I even went a fair ways towards implementing dual rust/ts import syntaxes. But saner people have pointed out that it's a little crazy to take on that maintenance burden forever if we can possibly avoid it.
> 
> 

I'm not suggesting that this project provide multiple syntaxes. But I would like to make sure the output of the linker is standardised so that different syntax can interoperate. For example the mangling scheme should be the same and metadata should be compatible

stefnotch:
The output of the linker isn't really intended to be used by other tools. As in, while I could import a generated `.wgsl` file in a `.wesl` module, I would have to deal with
- name mangling (could be standardised)
- eliminated dead code
- no visibility info
- no generics
...

Having the complexity of two syntaxes would be preferable over that IMO. But I do think that we're smart enough to figure out one syntax + semantics to go with it :)

PR #7, opened 2024-08-29
overlong argument for private by default in visibility discussion

mighdoll:


Visibility.md
```diff
@@ -84,4 +84,85 @@ If there is a file named `lib.wgsl` in the `package_root` directory,
...
+for default private at the package boundary. But on balance
+I think encouraging stronger internal apis is worth the cost of some syntactic noise.
+
+Note that public outside the package could be combined with public inside the package.
+So an element can be public to imports inside the package or public to imports both
+both inside and outside.
+
+### Visibility in Other Lanaguages
+
+* JavaScript and TypeScript are private by default, annotate for public.
```
k2d222, line 163: not completely, all members are public by default, opposite to rust. there's the _ prefix convention for private members, and the # standard prefix but for some reason, I have yet to see it in real code.

Python is similar, standard is double underscore but most ppl use a single one.

k2d222 (commented):
I like how you exposed these more philosophical questions behind our design choices. I feel in the long run we may want to move them to a design.md file to reduce noise in the doc. After all, most users just want to use stuff.

Since we have not yet standardized nor implemented anything, it's ok to have it here now.

possible later improvements:
* make it more concise
* citations and a section connecting to existing reflections. we are not the first to discuss these concepts.
* for completeness, section with arguments for public-by-default: e.g. private makes vanilla wgsl incompatible with wesl, pub everywhere adds noise, lib author sometimes forget a pub and users have to fork, weaker private alternatives exists, like underscore prefix, which can be hidden by language servers.

Would prefer using We instead of I.

Visibility.md
mighdoll: I was thinking of import export visibility in modules. (not object member visibility) 

https://www.typescriptlang.org/docs/handbook/2/modules.html

will make hthat more clear in the text

Visibility.md
```diff
@@ -84,4 +84,85 @@ If there is a file named `lib.wgsl` in the `package_root` directory,
...
+### Libraries Need API Control
+
+We want library publishers to be able to decide what to expose as
+part of their api. Libraries want to maintain and improve over time in a way
+that’s compatible with existing users. Otherwise publishers are stuck with their
+first implementation approach forever, or more likely publishers change the
+implementation anyway, and then library users need to expect to edit their wgsl
+code to upgrade to a new published library version. Upgrades get scarier for
+publishers and consumers. That added friction makes it harder to get to a viable
+wgsl library ecosystem. So I think we want to enable an api vs internal
```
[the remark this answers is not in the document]
mighdoll: will change.

Visibility.md
```diff
@@ -4,17 +4,23 @@
...
 * controlling host visible names like entry points and overrides?
 * Should export allow `as` renaming?
-* Treat .wgsl files as .wesl with every element exported?
 * Why not export struct Foo?
-  * Many current wgsl parsers (including wgpu's naga) would 
-    choke on the unknown attribute as is and feels like having 
+  * Many current wgsl parsers (including wgpu's naga) would
+    choke on the unknown attribute as is and feels like having
     two export forms is a bit inconsistent.
+* Consider changing to public by default for imports within the package only.
```
k2d222, line 14: I like this idea, especially with the prospect of importing wgsl/glsl etc. files locally and exposing them later at the package level :)

k2d222 (approved):
👍

Visibility.md
mighdoll: As I was drafting, I thought that the arguments for default private vs default public within the package were fairly even on both sides. I'm sure we'll discuss it further.

PR #8, opened 2024-08-29
Docs/goals-intro

mighdoll:
add goals and target programmer section to readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,29 +36,6 @@ or [discord](https://discord.gg/FXhZDV8V) to help.
 Our ambition is to create a modestly improved, practical variant of WGSL.
 The result should feel like WGSL with a few things added.
 
-## Enhanced WGSL Programmers
-
-Enhanced WGSL programmers are very likely to write shader programs
-only part time.
-And our programmers will learn WGSL first before learning our extensions.
-Our enhancements should be easy for our target programmers to learn
-and to remember.
-
-We imagine enhanced WGSL programmers will have learned WGSL,
-but not be familiar with other shader languages.
-They're very likely familiar with JavaScript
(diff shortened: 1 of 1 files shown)


k2d222 (commented):
Same comment as the other PR, it's fine here for now but when we have our shiny specs and impls to showcase, I would move these discussions to another file, for advanced users to read.

Our frontpage should quickly expose what we do, with links to more details.

README.md
```diff
@@ -31,15 +31,47 @@ A standard community library format is key to enable an WGSL library ecosystem.
...
+The target project size for enhanced WGSL ranges from tiny programs
+up to teams of few dozen programmers.
+Software engineering features useful for very large teams or very large
+projects are unlikely to be needed for enhanced WGSL.
+
 ## WGSL Enhancements
 
 We're aiming to prioritize WGSL enhancements that are: 1) important for community projects,
 2) feel natural to the WGSL programmer,
 and 3) reasonable to integrate into tools.
```
stefnotch, line 66: How about

"and 4) are simple to implement, helping widespread adoption"

mighdoll: maybe replace 3) with 'simple to integrate'? (having both 'reasonable to integrate' and 'simple to integrate')

mighdoll:
> Same comment as the other PR, it's fine here for now but when we have our shiny specs and impls to showcase, I would move these discussions to another file, for advanced users to read.
> 
> Our frontpage should quickly expose what we do, with links to more details.

I agree, doesn't belong here. will move to another file.

README.md
stefnotch: Sounds good

k2d222 (approved):

PR #9, opened 2024-08-30
Docs/designing-section

mighdoll:
doc section on principles for designing extensions

### DIFF
--- a/Designing.md
+++ b/Designing.md
@@ -83,10 +83,17 @@ so more projects can try them.
 We'll typically want to avoid using keywords used in current WGSL, unless
 the meaning is consistent and additive to the keyword's use in WGSL.
 
-We'll want to avoid conflicting with the meaning of keywords in future
-versions of WGSL as well.
-It's not yet clear whether we should avoid using all the keywords 
+Any differences we introduce in WESL will
+will be erased during transpilation to WGSL.
+But we don't want to confuse our programmers by
+unexpectedly changing the meaning of a keyword.
+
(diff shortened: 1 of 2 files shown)


k2d222 (approved):
Good overview of our design choices for new contributors 👍

Designing.md
```diff
@@ -0,0 +1,92 @@
...
+As experiments move toward stability
+we'll move experimental enhancements into more tools behind a feature flag
+so more projects can try them.
+
+## Words and Reserved Words in WGSL
+
+We'll typically want to avoid using keywords used in current WGSL, unless
+the meaning is consistent and additive to the keyword's use in WGSL.
+
+We'll want to avoid conflicting with the meaning of keywords in future
```
stefnotch, line 86: We could get away without this requirement.

1. Whenever WGSL itself gets a new syntactical construct, all the WESL tools need to be updated anyways
2. With the editions mechanism, it is possible to completely change WESL without breaking any of the consumers

We probably still want at least a relaxed version of this.

mighdoll: we'll have the technical means to translate anything w/o conflict, I agree!

The remaining question is what to do if there's a likely future confusion.. On the one hand, if we're prototyping a wgsl feature than might just get adopted as is, it'd be better for everyone if we use a reserved word. But if we're expecting the meaning of the keyword to be different in a future rev, then better to use a different word.
  
I'll try an edit..

stefnotch: That sounds like a very reasonable option, I approve of this.

PR #11, opened 2024-08-31
Main

mighdoll:
mention discussion of conditions for build time linking


k2d222 (approved):
Adding to the discussion:

Should we support both runtime and static linking at the same time, like unity does? 

Can runtime linking relax some requirements for conditionals, since checking does not have to be as strict? e.g. could conditionals depend on pipeline-overridable consts? 

How do we choose what linking strategy to use in what cases? Can the programmer choose, is it transparent?

From an implementation perspective, how to deal with multiple shader variants since they would be in different files but have the same names and contents. How to import them etc.

mighdoll:
re: static linking. It'd be nice to work with some projects that want to pre-compile shader variants to understand their practical issues.

PR #18, opened 2024-09-02
Update GLOSSARY.md

stefnotch:
- Changes the name to WESL
- Adds the term "root module" to the glossary, since we've been using that on Discord.


mighdoll:
LGTM.

Also, just realized that WESL also is short for 'We Shading Language'. i.e. a community shading language

PR #19, opened 2024-09-05
attributes-based conditional compilation proposal

k2d222:


ConditionalCompilation.md
```diff
@@ -1,28 +1,140 @@
...
+| unstructured | structured |
+|--------------|------------|
+| (+) much easier to implement: just look for the `#word` symbols | (-) harder to implement: requires a full parsing |
+| (+) often language-agnostic (see the [C preprocessor](https://en.wikipedia.org/wiki/C_preprocessor). Familiar to C developers. | (-) a new syntax needs to be taught, not always self-explanatory |
+| (-) poorly integrated in the language, harder to read by humans | (+) is a well-designed syntax feature of the language |
+| (+) technically more expressive (e.g. [manipulating identifiers](https://en.wikipedia.org/wiki/C_preprocessor#Token_concatenation)) | (-) can only conditionaly include parts of the syntax tree, poor text generation capabilities. |
+| (-) behaves unpredictably, intent and behavior is hidden | (+) intent and behavior is visible at the usage site |
+| (-) no type checking, no syntax checking, because code is generated dynamically | (+) can statically check that conditional variants lead to syntactically correct and type-valid code |
+| (-) poor IDE support | (+) IDE can check all possible code paths |
+
```
mighdoll, line 22: Excellent discussion. Though (as you reminded me on a recent pr!) I think it should move to a separate location so folks can visit the spec and not have to read the back story unless they’re interested.

ConditionalCompilation.md
```diff
@@ -1,28 +1,140 @@
...
+```rs
+@ifdef(my_feature)
+fn f() { ... }
+@ifndef(my_feature)
+fn f() { ... }
+```
+
+### Syntax 2: `@if` attribute with compile-time logical expressions
+
```
mighdoll, line 78: Syntax 1 and Syntax 2 are alternate, or is the idea to do both?

Given what people are doing, I suspect we need the flexibility of multiple conditions.

ConditionalCompilation.md
```diff
@@ -1,28 +1,140 @@
...
+*Example*
+
+```rs
+@if(feature_1 && (!feature_2 || feature_3))
+fn f() { ... }
+@elif(!feature_1)
+fn f() { ... }
+@else
+fn f() { ... }
+```
```
mighdoll, line 102: hmm. nice that it's familiar to users of #if #else #endif. 
but also doesn't add much vs. #if #else #endif...

ConditionalCompilation.md
mighdoll: What enables the IDE to know the valid combinations of conditions? 

And if there is a way to know the valid combinations of conditions, why is the siutation different for structured vs. unstructured?

ConditionalCompilation.md
```diff
@@ -1,28 +1,140 @@
...
+
+### Definitions
+
+ * **Compile-time expression**: A *compile-time expression* is evaluated by the compiler before the rest of the program. It lives in the *compile-time scope*.
+
+ * **Compile-time scope**: The *compile-time scope* this is an independent scope from the [*module scope*](https://www.w3.org/TR/WGSL/#module-scope), meaning it cannot see any declarations from the source code, and its identifers are independent.
+
+ * **Compile-time feature**: A *compile-time feature* is an identifier that evaluates to a boolean. It is set to `true` if the feature is *enabled* during the compilation phase.
+
+ * **Compile-time attribute**: A *compile-time attribute* is parametrized by a *compile-time expression*. It is eliminated after the compilation phase but can affect the syntax node it decorates.
```
mighdoll, line 36: How do we handle runtime conditions?

ConditionalCompilation.md
```diff
@@ -1,28 +1,140 @@
...
+fn response_to_the_ultimate_question() -> u32 {
+  return 42;
+}
+
+@if(response_to_the_ultimate_question() == 42)
+fn f() { ... }
+```
+
+* Decorating other WESL language extensions: import statements could be decorated with *compile-time attributes* too.
+
```
mighdoll, line 132: I was going to ask about that! I think some projects are doing this in the wild, but I don't know whether its a needed mechanism or if there are alternatives.

ConditionalCompilation.md
```diff
@@ -1,28 +1,140 @@
...
+A Compile-time attribute CAN decorate the following syntax nodes:
+ * structure declarations
+ * structure members
+ * function declarations
+ * ?? function formal parameter declarations ?? (is this needed?)
+ * variable declarations and value declarations
+ * all statements, except those disallowed below
+ * directives
+
+Because eliminating the decorated node would lead to invalid code, a Compile-time attribute CANNOT decorate the following syntax nodes:
```
mighdoll, line 51: projects are doing some of these things with homegrown ifdef implementations

do we have an alternate solution for them? e.g. perhaps our other features may cover for their needs? or perhaps you've got a runtime condition mechanism in mind?

mighdoll (commented):
The need for conditional compilation is clear. Lots of projects have found the need to build it themselves.

mighdoll (commented):
arguing for #ifdef's merits is a way to explore the issues, though I'm hoping we find something better.

mighdoll (commented):
Its really helpful that you've dug into this, there was more complexity than I realized.

ConditionalCompilation.md
k2d222: valid combinations cannot be checked. This is the case e.g. in rust too.
I'll clarify this.

ConditionalCompilation.md
k2d222: I think I'll just keep syntax 2. Before writing this I wasn't sure about this so I wrote both. But the idea was to keep one.

k2d222: because at first syntax 2 seemed too complicated, but then I realized that we could just support a simple syntax consisiting of idents, &&, || and !. That can be expanded in thefuture.

ConditionalCompilation.md
k2d222: Discord discussion: 
```
say we invoke the compiler/linker this way:

const features = { feat1: true, feat2: false, feat3: undefined }
const compiled = linker.link(source, features)

then this file:

@if(feat1)
fn func1() {}

@if(feat2)
fn func2() {}

@if(feat1 && feat3)
@if func3() {}

@if(feat2 && feat3)
@if func4() {}


would be compiled like this:

fn func1() {}

// func2 is eliminated

@if(true && feat3) // attribute partially resolved
@if func3() {}

@if(false && feat3) // attribute partially resolved
@if func4() {}
```

I'll add this to the spec in the execution section.

ConditionalCompilation.md
stefnotch: To answer that: With a structured `@if`, one can always generate a unified syntax tree

```
struct Foo {
  a: f32,
  @if(some_condition)
  b: f32
}
```
would turn into a syntax tree along the lines of
```
(struct foo (member a) (if some_condition (member b)))
```

This means that at the very least, a linker can always check if the WGSL code is syntactically (not semantically) valid.

Meanwhile with `#ifdef`, one can frequently end up with multiple separate syntax trees. The above example could be turned into a single syntax tree. Meanwhile
```
struct FooBar { // Which bracket is the closing bracket?
  a: f32,
#ifdef some_condition
}
#else
  b: f32
}
#endif
```
is *very* hard to turn into a single syntax tree. Juggling two separate syntax trees in a linker or language server is a whole lot of extra effort, so nobody does it.
This example in particular also breaks one IDE feature: Jump-to-matching-bracket


And finally, `@if` is equally powerful as `#ifdef`. Proof:
- Assume all combinations of conditions are valid
- Expand the `#ifdef` into all of the combinations
- Encode each combination in a separate `@if`

The above `FooBar` example could thus be naively translated as
```
@if(!some_condition)
struct FooBar { 
  a: f32,
}
@if(some_condition)
struct FooBar { 
  a: f32,
  b: f32
}
// Oh look, two syntactically valid WGSL structs. Happiness :)
```
Yes, that's code duplication, but it's this example is just there to illustrate the fact that `@if` is equally powerful.

ConditionalCompilation.md
stefnotch: IMO we don't really need `@elif` and `@else` in the simplest variant of the proposal. I believe they also complicate the parsing a fair bit, since the following should be illegal.
```
@if(feature_1)
struct Foo { a: f32}
struct Bar { a: f32 }
@else
struct Foo { a: vec3f }
```

ConditionalCompilation.md
stefnotch: "runtime conditions" as in "we run the compiler at runtime, instead of having an offline compilation pass"?

And not "runtime" as in "runtime of the shader", right?

ConditionalCompilation.md
stefnotch: It's a reasonably useful mechanism for say, having a specialized mobile GPU shader. (target platform)
Then one doesn't have to put a hundred `@if`s into their mobile GPU shader, but can instead say "only bother importing if we are on mobile".

ConditionalCompilation.md
stefnotch: I don't fully understand why decorating all of these nodes would lead to invalid code?

As in, this could work. It'd just eliminate the entire if-else statement
```
@if(feature_1)
if (a<b) {
...
} else {
...
}
```
Meanwhile that could reasonably be disallowed
```
if (a<b) {
...
} 
@if(feature_1)
else {
...
}
```
Because people are supposed to write
```
if (a<b) {
...
} else {
@if(feature_1) {
...
}
}
```

ConditionalCompilation.md
k2d222: Not sure about this too and didn't give a lot of thought. So I'll keep it for future reflexion.

One constraint it would add is that the conditional compilation pass would **have to** run before import resolution.
But that does not sound like a big concern.

stefnotch: Regarding the `@comptime`: Very useful, but sufficiently out of scope for the time being. I'd say that the alternative is letting those projects evaluate such conditions themselves in Javascript/Rust code. And that injecting code into shaders by using the linker's APIs (which are not part of the specification) is always an option.

k2d222:
> Technically there's also the option of integrating into the WGSL compile time features. As in, WGSL has features like [`const_assert`](https://www.w3.org/TR/WGSL/#const-assert-statement) which needs a const expression. And that expression is evaluated at shader creation time. Theoretically those could be evaluated beforehand by a smart-enough linker.

yes, That's why I mentionned the `@comptime` example at the end. I think requiring that linkers support full evalutation of const expressions is a huge implementation burden. They would have to evaluate const functions too. In any case, I think V1 can use that simpler expression grammar and see if we need to expand it. But I doubt that.

and I think that's something we have not discussed too much so far: do we want our spec to be modular? e.g. can users choose if they want just the import feature or just the cond-comp? because right now, they could behave independently.
If a user enabled both the comptime extension and the cond-comp extension, then they would benefit from richer expressions in the @if attributes.

ConditionalCompilation.md
k2d222: What are you referring to?

stefnotch: > Because eliminating the decorated node would lead to invalid code, a Compile-time attribute CANNOT decorate the following syntax nodes:
>  * bodies of function declarations, switch statements, switch clauses, loop statements, for statements, while statements, if/else statements
> * function return types

stefnotch: Is that just because of how the WGSL grammar is set up?

ConditionalCompilation.md
stefnotch: Sounds good!

ConditionalCompilation.md
k2d222: @stefnotch that would not be caught at parse time, more likely as a later contextual check. 
but yeah I agree that elif else are not necessary at first. That was also part of the idea of proposal 1 with only ifdef and ifndef.

I'd say we can see how hard it is to implement them. I have the feeling that it's not too much additional work to support them.

ConditionalCompilation.md
k2d222: I understood the former, i.e. for example a feature might depend on a runtime condition (e.g. supported hardware). In this case one has two choices: generate all possible shader variants or compile at runtime (the start of the program).

We could do a mix of that, generating a partially compiled wesl shader (all imports resolved, most of the cond-comp resolved) and run a cheap 2nd compilation with the missing featuers.

ConditionalCompilation.md
k2d222: I will clarify this. But I refered to the **bodies**, because the grammar for these is like this:
`attribute* if/else/while... () CompoundStatement`, where `compoundStatent` is the body and looks like `attribute* { Statement* }`

so you don't want to have
`while (true) @if(false) {}`

because the CompoundStatement is not optional.

k2d222: > Meanwhile that could reasonably be disallowed
> 
> ```
> if (a<b) {
> ...
> } 
> @if(feature_1)
> else {
> ...
> }
> ```
> 
> Because people are supposed to write
> 
> ```
> if (a<b) {
> ...
> } else {
> @if(feature_1) {
> ...
> }
> }
> ```

Why? your first variant looks more correct to me. Should you not be able to erase an `else`?

stefnotch: I see, I read it as 

> (bodies of function declarations), switch statements, switch clauses, loop statements, for statements, while statements, if/else statements

instead of the intended

> bodies of (function declarations, switch statements, switch clauses, loop statements, for statements, while statements, if/else statements)

> Should you not be able to erase an `else`?

The WGSL grammar doesn't allow an attribute in that position. Rust also disallows attributes in that position.

k2d222: > The WGSL grammar doesn't allow an attribute in that position. Rust also disallows attributes in that position.

Oh I missed that one, thanks! about rust: in rust if/else can be an expression whereas if alone cannot. In wgsl, this is not a problem as if/else cannot be and expression. So I suggest to add else to the list of places where we would allow attributes.

ConditionalCompilation.md
mighdoll: +1 for providing a proof

ConditionalCompilationDesign.md
```diff
@@ -0,0 +1,66 @@
+> Corresponding spec: [ConditionalCompilation.md](ConditionalCompilation.md)
+
+## Initial reflexions
```
mighdoll, line 3: reflections

ConditionalCompilation.md
mighdoll: +1 for starting simple if it seems sufficient to you.

ConditionalCompilation.md
mighdoll: Exactly - I meant running the compiler at runtime vs build time.

While partially compiling at build time and then completing the compile at runtime is possible, I'd suggest we skip mixed mode until we need it.

ConditionalCompilation.md
```diff
@@ -1,28 +1,139 @@
...
+  @if(legacy_implementation || (is_web_version && xyz_not_supported))
+  let result = legacy_impl();
+  @else
+  let result = modern_impl();
+}
+```
+
+## Definitions
+
+ * **Compile-time expression**: A *compile-time expression* is evaluated by the compiler before the rest of the program. It lives in the *compile-time scope*.
```
mighdoll, line 39: I was a little confused by the phrase 'compile-time' (as you saw in my earlier comments). I hear 'compile time' and think that must be running at build time, and so settings will be fixed at runtime (e.g. the gpu capabilities). Of course here compilation may happen at runtime, instead of at build time. 

Lest others be confused, I wonder if we might use a different word than compile-time. transpile-time, assembly-time, perhaps? Or if there's no good alternative phrase, then I'd suggest an explanation right up front that compilation can happen at runtime.

ConditionalCompilation.md
```diff
@@ -1,28 +1,139 @@
...
+
+Conditional compilation is a mechanism to modify the source code based on parameters passed to the compiler.
+This specificaton extends the [*attribute* syntax](https://www.w3.org/TR/WGSL/#attributes) with three new attributes: `@if`, `@elif`, `@else`.
+These attributes indicate that the syntax node they decorate can be removed by the compiler based on feature flags.
+
+> Note: This implementation is similar to the `#[cfg(feature = "")]` syntax in rust.
+
+### Usage Example
+
+```rs
```
mighdoll, line 15: love the examples! 

(also, reminder to drop @else from example if dropping @else)

mighdoll (commented):
I suggest we do a jog through the community projects we know of to see if their existing use of conditional compile / conditional assembly poses any challenges to the scheme. (I don't think that needs to hold up this PR though)

ConditionalCompilation.md
```diff
@@ -1,28 +1,139 @@
...
+    * If the feature is *enabled*, the identifier is replaced with `true`.
+    * If the feature is *disabled*, the identifier is replaced with `false`.
+   
+   The expression is *fully resolved* if all its *compile-time features* have been replaced by `true` or `false`.
+5. *Compile-time attributes* which have *fully resolved expressions* are evaluated:
+    * If the syntax node is marked for removal: it is eliminated from the source code along with the attribute.
+    * Otherwise, if all features in only the attribute is eliminated from the source code.
+4. The updated source is passed to the next compilation phase. (e.g. import resolution)
+
+## Possible future extensions
```
mighdoll, line 112: Some old discussion suggested that default conditions ought to be set in the wgsl, for IDEs. WDYT? 

I suppose that could be handled subsequently to this PR, but I suggest adding it to the possible future section.

ConditionalCompilation.md
stefnotch: "Translation time"? And we just define that word up front.

mighdoll: +1 for translation time!

(but agree that explaining up front is sufficient if no alternate phrase appeals)

k2d222: OK for me for translation time. could add that to the glossary too. Also, I never know if we should be using linker/compiler/transpiler/preprocessor... I think it would be favorable to decide that, as the term will appear in the spec. My preference is compiler.

k2d222: Renamed compilation to translation and renamed compiler to WESL translator.
Not a big fan of WESL translator, but I chose it for consistency.

Open to bikeshedding :)

ConditionalCompilation.md
k2d222: I don't see an imediate need for this, and I don't see an elegant syntax for it, so I suggest we defer but keep in mind.

k2d222:
Ready for further review!

I would like to settle on a term for the WESL compiler/linker/preprocessor/translator/...

I prefer compiler. It is the most self-explanatory in my eyes.
* linker could only refer to the linking phase aka. resolving imports
* preprocessor seems too limiting for the current scope of WESL
* translator is cool, but not really standard, and "conditional translation" does not sound good.
* Transpiler could be an option, but it is not a familiar term to many developers.

ConditionalTranslation.md
```diff
@@ -0,0 +1,297 @@
...
+fn f() { ... }
+```
+
+## Execution of the conditional translation phase
+
+The conditional translation phase *should* be the first phase to run in the full WESL-to-WGSL translation pipeline.
+
+> NOTE: The conditional translation was designed to be incremental. In case some features can only be resolved at runtime, the *WESL translator* can be invoked in multiple passes:
+> A first pass is invoked with features determined at compile-time and returns a partially translated WESL source.
+> A second pass is invoked with features determined at run-runtime and returns a fully translated WGSL source.
```
stefnotch, line 122: Damn, I wish we had better names for that. Right now I can't think of a significantly better option though.

stefnotch (approved):
This is excellent, I believe I have no further concerns.

ConditionalTranslation.md
```diff
@@ -0,0 +1,297 @@
+# Conditional Translation
+
+## Overview
+
+> This section is non-normative
+
+Conditional translation is a mechanism to modify the output source code based on parameters passed to the *WESL translator*.
```
mighdoll, line 7: 'conditional translation' sounds clear.

ConditionalTranslation.md
```diff
@@ -0,0 +1,297 @@
...
+
+### Summary
+
+1. The grammar is extended to allow *translate-time attributes* before the following syntax nodes:
+   * const value declarations
+   * variable declarations
+   * directives
+   * switch clauses
+   * else if clauses and else clauses
+   * all statements that did not allow attributes:
```
mighdoll, line 65: 'all statements that did not allow attributes' is a little confusing.. I suspect you're distinguishing statements from expressions and declarations..

ConditionalTranslation.md
```diff
@@ -0,0 +1,297 @@
...
+* **Translate-time scope**: The *translate-time scope* this is an independent scope from the [*module scope*](https://www.w3.org/TR/WGSL/#module-scope), meaning it cannot see any declarations from the source code, and its identifers are independent.
+
+* **Translate-time feature**: A *translate-time feature* is an identifier that evaluates to a boolean. It is set to `true` if the feature is *enabled* during the translation phase and `false` if the feature is *disabled*.
+
+* **Translate-time attribute**: A *translate-time attribute* is parametrized by a *translate-time expression*. It is eliminated after the translation but can affect the syntax node it decorates.
+
+## Location of *Translate-time attributes*
+
+A *translate-time attribute* can appear in all places where the WGSL grammar allows attributes, except if removal of the decorated syntax node would lead to syntactically incorrect code.
+The grammar is extended to allow attributes in several locations previously not allowed by the WGSL grammar. These extensions are specified in section [updated grammar](#updated-grammar).
```
mighdoll, line 55: can you summarize the attributable locations for the programmer to understand? The precise details are important to specify, but it's a long list to expect the programmer to remember.

ConditionalTranslation.md
```diff
@@ -0,0 +1,297 @@
...
+
+### Summary
```
mighdoll, line 57: It is nice to be able to condition almost anything, but it's a big change both to the grammar and to programmer understanding of what can be conditioned. 

Is there a smaller change that would be almost as good? What if there were only a few things that could be conditioned, e.g. conditioning top level items is easy to understand: functions, structs, directives. (although then it gets tempting to allow conditioning struct members, and maybe function calls.. I see that it's slippery slope)

ConditionalTranslation.md
```diff
@@ -0,0 +1,297 @@
...
+3. A *translate-time attribute* CANNOT decorate the following syntax nodes, even if the WGSL grammar allows attributes before these syntax nodes:
+   * function return types
+   * the body (part surrounded by curly braces) of:
+     * function declarations
+     * switch statements
+     * switch clauses
+     * loop statements
+     * for statements
+     * while statements
+     * if/else statements
```
mighdoll, line 95: confusing that you can condition `else if` but not `if`..

mighdoll:
what about import statements? (I don't recall where, but I think I've seen wgsl projects already conditioning their home grown imports with #ifdef.)

mighdoll:
Why can't you condition loops and if statements? It seems like the kind of thing programmers would want to condition if condition anything is the mantra.

mighdoll:
Can you discuss how language servers will handle this feature (perhaps in the appendix?)

mighdoll:
What happens with punctuation, here the semicolon?
`let result = @if(legacy_mode) legacy_impl() @else modern_impl();`

ConditionalTranslation.md
```diff
@@ -0,0 +1,297 @@
...
+@if(feature_1 && (!feature_2 || feature_3))
+fn f() { ... }
+@if(!feature_1)                               // corresponds to @elif(!feature_1)
+fn f() { ... }
+@if(feature_1 && !(!feature_2 || feature_3))  // corresponds to @else
+fn f() { ... }
+```
+
+## Execution of the conditional translation phase
+
```
mighdoll, line 117: Explain just what the wgsl programmer needs to understand about the execution model here.
I think the rest is implementation advice for linker implementators that the wesl user can ignore, and so can go in the appendix.

ConditionalTranslation.md
```diff
@@ -0,0 +1,297 @@
...
+The conditional translation phase *should* be the first phase to run in the full WESL-to-WGSL translation pipeline.
+
+> NOTE: The conditional translation was designed to be incremental. In case some features can only be resolved at runtime, the *WESL translator* can be invoked in multiple passes:
```
mighdoll, line 120: minor wording comment: 'incremental' implies many small, so probably not the right word here. I think you mean there might be two passes, one at build time and one at runtime..

mighdoll:
is 0 false? I imagine some would like to do `@if(num_textures)`

ConditionalTranslation.md
k2d222: yes, I am referring to the '*Statement' syntax nodes in the grammar. I'll clarify this.

ConditionalTranslation.md
k2d222: :+1: can do

mighdoll:
Summarizing my comments:
1) Some presentation changes for the wesl programmer as reader, e.g. wording and separating some of the details to the appendix
2) I agree that simple is good! The syntax is clear and fits in well, and annotations will be nicer than ifdef. My main remaining concern is that the rules for what can be conditioned will feel complex. WDYT? If conditioning fewer things were feasible that seems like it'd be nice place to start, if not, can you explain more why can't we condition everything (including e.g. if statements and loops)?

ConditionalTranslation.md
k2d222: to the grammar, it's honestly not such a big change. just added `attribute*` before all declarations, statements, the struct members, switch clauses and the else/else if clauses. It does look scary in the spec though.

To the programmer, it's quite easy: you can place them anywhere where it would make sense to have an conditional attribute. It's very similar to decorators in typescript, annotations in java, and attributes in rust. In my experience I never learned where I could use them for these languages, I just assumed I could.

ConditionalTranslation.md
k2d222: you can condition if, but it's already in the grammar. That's why it's not mentionned here.

Overall seeing your comments I need to rephrase that this section refers to edits to the grammar. I'll add a more human-friendly explanation too.

ConditionalTranslation.md
k2d222: The spec is quite formal, not really targeted at end-users. I'll write a friendlier doc when it's merged.

ConditionalTranslation.md
```diff
@@ -0,0 +1,297 @@
...
+3. The *translate-time features* in *translate-time expressions* are resolved:
+   * If the feature is not present in the *WESL translator*'s feature list, it is unchanged.
+   * If the feature is *enabled*, the identifier is replaced with `true`.
+   * If the feature is *disabled*, the identifier is replaced with `false`.
+   
+   The expression is *fully resolved* if all its *translate-time features* have been replaced by `true` or `false`.
+
+4. *Translate-time attributes* which have *fully resolved expressions* are evaluated:
+   * If the decorated syntax node is marked for removal: it is eliminated from the source code along with the attribute.
+   * Otherwise, only the attribute is eliminated from the source code.
```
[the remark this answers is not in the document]
k2d222: agreed, and perhaps it doesn't have to be normed at all? The implementations are free to do it anyway. It can be added as a non-normative comment.

k2d222:
@mighdoll answering the rest of your comments,

> what about import statements? (I don't recall where, but I think I've seen wgsl projects already conditioning their home grown imports with #ifdef.)

I would love to support conditional compilation on import statements. They do not require any extra work on this spec. They are already mentionned in the **Possible future extensions** section.

> Why can't you condition loops and if statements? It seems like the kind of thing programmers would want to condition if condition anything is the mantra.

That section was visibly too confusing. It only refers to the changes to the LR or recursive-descent grammar, but it is redundant anyway with the appendix. So I'll rewrite it in a more human-friendly way :) thanks for the heads up, it clearly needs a rewrite :+1: 

> Can you discuss how language servers will handle this feature (perhaps in the appendix?)

Aah, yeah very good point. Let's merge this first? Would need to think a lot about this before I could just write it down.

> What happens with punctuation, here the semicolon? `let result = @if(legacy_mode) legacy_impl() @else modern_impl();`

Here you're decorating a `CallExpression`, and all `Expression`s (as of this proposal) cannot be decorated with attributes. The reason is that they are not optional in the general case, so removing them leads to invalid syntax. I made sure that all attributes always lead to correct syntax no matter the features.

But I also mentionned in the  **Possible future extensions** section that with the `@else` attribute, we could (not saying should) allow attributes on `Expression`s as well.
 
> is 0 false? I imagine some would like to do `@if(num_textures)`

Good question! In the current proposal, only feature flags (boolean) can be accepted in the limited grammar of `@if` expressions. So, `num_texture` would not be allowed here in the first place. In section **Possible future extensions** I discuss the possibility of enlarging this syntax. I kept it simple because otherwise it requires to write a full WGSL *interpreter* of expressions to run them at translate-time.

But even if it was allowed, you would have to write `@if(num_texture == 0u)` explicitly. This is by design: WGSL has no implicit conversion to boolean. In fact [it only has implicit conversions from abstract to concrete numeric types](https://www.w3.org/TR/WGSL/#conversion-rank) (and composite types of them).

k2d222:
re: making it clearer for the human reader: I think this spec lacks a nice, self-explanatory example. A reader could just look at it and understand without diving into the nitty details I'm sure. It's quite straightforward if you're familiar with attributes and conditional compilation.

ConditionalTranslation.md
mighdoll: I'm glad it'll be easy! It think I'm not used to being able to put decorators/annotations on as many things and misread this as I was trying to understand where they could apply.

mighdoll:
> > Why can't you condition loops and if statements? It seems like the kind of thing programmers would want to condition if condition anything is the mantra.
> 
> That section was visibly too confusing. It only refers to the changes to the LR or recursive-descent grammar, but it is redundant anyway with the appendix. So I'll rewrite it in a more human-friendly way :) thanks for the heads up, it clearly needs a rewrite 👍

I'm sorry I didn't read it with more care to catch your intent. At least my misunderstanding will help to clear things up for the many readers to come. 

> > Can you discuss how language servers will handle this feature (perhaps in the appendix?)
> 
> Aah, yeah very good point. Let's merge this first? Would need to think a lot about this before I could just write it down.

yep, later is fine.

later is fine too for more complete user level docs. 

> > What happens with punctuation, here the semicolon? `let result = @if(legacy_mode) legacy_impl() @else modern_impl();`
> 
> Here you're decorating a `CallExpression`, and all `Expression`s (as of this proposal) cannot be decorated with attributes. 

I think it'll really help me to understand the list of which things can and can't be decorated, without having to cross reference to the grammar.

> The reason is that they are not optional in the general case, so removing them leads to invalid syntax. I made sure that all attributes always lead to correct syntax no matter the features.

Why is that necessary? Even if the syntax works, the code might still quickly fail in the dawn/wgpu compiler, e.g. if you conditionally disable a var, but not the reference to the var.

> 
> > is 0 false? I imagine some would like to do `@if(num_textures)`
> 
> Good question! In the current proposal, only feature flags (boolean) can be accepted in the limited grammar of `@if` expressions. So, `num_texture` would not be allowed here in the first place. In section **Possible future extensions** I discuss the possibility of enlarging this syntax. I kept it simple because otherwise it requires to write a full WGSL _interpreter_ of expressions to run them at translate-time.

> 
> But even if it was allowed, you would have to write `@if(num_texture == 0u)` explicitly. This is by design: WGSL has no implicit conversion to boolean. In fact [it only has implicit conversions from abstract to concrete numeric types](https://www.w3.org/TR/WGSL/#conversion-rank) (and composite types of them).


that seems a reasonable rule, since the attribute vars look like wgsl booleans, they should behave like wgsl booleans (and not preprocessor vars)

k2d222:
> Why is that necessary? Even if the syntax works, the code might still quickly fail in the dawn/wgpu compiler, e.g. if you conditionally disable a var, but not the reference to the var.

It's a fair question, I guess it avoids one possible source of bugs, it's already something. It's the best we can reasonably do.

Breaking the syntax also leads to the issue @stefnotch mentionned earlier with the #ifdef: the linters cannot construct a syntax tree encompassing all possible configurations.

mighdoll (approved):
LGTM

PR #34, opened 2024-09-19
🚧 Generics proposal

k2d222:
**⚠️ This is a historical proposal kept open for reference in discussions. ⚠️**

A proposal for function and modules generics using respectively the `@generic` and `@override` attributes.

Still rough around the edges, need more thought. Ideas and comments are welcome.

### DIFF
--- a/Generics.md
+++ b/Generics.md
@@ -1,6 +1,6 @@
 # Generic Declarations and Generic Modules
 
-I would like to reduce to the maximum the added syntax and semantics in our extensions.
+I would like to limit as much as possible the amount of additional syntax and semantics in our extensions.
 After all, we are writing an extension, not a different language!
 It should look and feel just like the original stuff and extensions should be straight forward, self-explanatory even.
 
@@ -14,9 +14,9 @@ It builds on 3 observations
 * global value and alias declarations are immutable and must be initialized. What if we could override them? if so, this is a simple mechanism for module specialization.
   What if we also could override struct and functions declarations? In fact, all declarations?
 
-It proposes two extensions: **Function Generics** and **Module Generics** (+ an optional **Module-as-Types**).
-They are orthogonal bu they share the concepts of *Type Hierarchy* and *Type Constraints* formalized below.
-We could support one of the two and achieve the same expressivity AFAIK. But not the same ergonomics.
+It proposes two extensions: **Function Generics** and **Module Generics** (+ optional **Function-as-Type** **Module-as-Type**).
+They are orthogonal but they share the concepts of *Type Hierarchy* and *Type Constraints* formalized below.
+We could support only one of the two and achieve the same expressivity AFAIK. But not the same ergonomics.
 
 ## Type Hierarchy
 
@@ -87,7 +87,7 @@ Note that compared to the WGSL spec, we need to constrain a bit more what can be
 So obviously, this level of constraining is too high to express in the grammar.
 
 TL;DR instead, we could specify these constraints:
-* Unconstrained.
+* Unconstrained (or implicitly constrained by [AllTypes](https://www.w3.org/TR/WGSL/#alltypes-type).
 * Constrained by: a user-defined struct, `vec`, `mat`, `array`, `atomic`, `ptr`, `AbstractInt` or `AbstractFloat`: Must be specialized by a *constructible* subtype.
 * Constrained by a leaf type: `u32`, `i32`, `f32`, `f16`, `bool`: Must be specialized with a instance of that type (like the `N` parameter).
 
@@ -120,6 +120,35 @@ fn rotate(point: Point2D<T>, angle: f32) -> Point2D<T> {}
 
 alias mat4x4f = matrix<f32, 4, 4>;
 ```
+
+## (optional) Function-as-Type
+
+Since a function signature now has a well-defined type and subtyping, we *could* support "function pointers" via genericity.
+
+You cannot pass by value a function pointer, instead it is passed as a specialization of the generic parameter.
+
+This can be implemented simply by substitution of the generic type, like the rest of the generics. No extra work needed except for the type-checking.
+
+```rs
+@generic(T)
+fn callback_signature(arg: T) -> f32; // just define an abstract function, this serves only the purpose of expliciting the signature.
+
+@generic(T, F: callback_signature<T>)
+fn my_fn() -> f32 {
+  // ...
+  return F(val2); // call the generic function (similar to a function pointer).
+}
+
+fn callback_implementor(arg: u32) -> f32 { // this function signature is a subtype of callback_signature
+  // ...
+}
+
+fn main() {
+  my_fn<u32, callback_implementor>();
+}
+```
+
+See [the map reduce example](#map-reduce).
 
 ## Generic Modules
 
@@ -194,7 +223,7 @@ const linkerOptions = {
 let wgsl_source = linker.link("main.wesl", linkerOptions)
 ```
 
-## (optional) Module-as-Types
+## (optional) Module-as-Type
 
 We can allow a module to be passed as a type constraint. Then, a declaration with a module generic can be specialized with any other module that overrides all uninitialized declarations. This can help implement object-oriented or interface semantics.
 
@@ -255,7 +284,9 @@ A compile-time linker or a IDE linter/langserver should implement type checking.
 
 ## Examples
 
-This example uses function generics, module generics and module-as-types.
+### Object Oriented
+
+This example uses function generics, module generics and module-as-type.
 
 ```rs
 // file shape.rs: it acts as an interface (or a base class in OOP).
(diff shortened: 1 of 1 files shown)


Generics.md
```diff
@@ -1,164 +1,386 @@
...
+`array` is parameterized by `T` and `N`. `arrayLength` is parameterized by `E` and `AM`.
+* `T` and `E` must be one of: a `scalar`, `vector`, `matrix`, `atomic`, `array`, structure[^1].
+* `N` must be an override-expression that evaluates to a `i32` or `u32`, and greater than 0[^2].
+* `AM` must be a value of enumerant [`access mode`](https://www.w3.org/TR/WGSL/#access-mode)[^3].
+
+So obviously, this level of constraining is too high to express in the grammar.
+
+TL;DR instead, we could specify these constraints:
+* Unconstrained (or implicitly constrained by [AllTypes](https://www.w3.org/TR/WGSL/#alltypes-type).
+* Constrained by: a user-defined struct, `vec`, `mat`, `array`, `atomic`, `ptr`, `AbstractInt` or `AbstractFloat`: Must be specialized by a *constructible* subtype.
```
ncthbrt, line 93: How would a user defined struct constraint work?

Generics.md
```diff
@@ -1,164 +1,386 @@
...
+* `VecN<S>` is a subtype of `VecN<T>` if `S` is a subtype of `T`
+* `MatCxR<S>` is a subtype of `MatCxR<T>` if `S` is a subtype of `T`
+* `Array<S,N>` is a subtype of `Array<T,N>` if `S` is a subtype of `T`
+
+And we can add to that list:
+* `atomic<S>` is a subtype of `atomic<T>` if `S` is a subtype of `T`.
+* `ptr<S>` is a subtype of `ptr<T>` if `S` is a subtype of `T`.
+
+We can extend subtyping for user-defined types:
+* structure type `S1` is a subtype of structure type `S2` if, for all members `m` in `S2`: `S1.m` exists and `S1.m` is a subtype of `S2.m` (member covariance).
```
ncthbrt, line 37: This is certainly an interesting property, but also feels a bit dangerous because of how  memory layout would differ, potentially impacting stuff like arrays, vecs, etc

Generics.md
```diff
@@ -1,164 +1,386 @@
...
+
+It builds on 3 observations
+
+* WGSL *already* has support for parametric types and functions, they are called [type-generators](https://www.w3.org/TR/WGSL/#type-generator) and [function overloads](https://www.w3.org/TR/WGSL/#builtin-functions). It even has some syntax for them.
+  The issue is, they are not allowed to be declared in user code. This proposal changes this.
+* A *module* is a collection of functions, structs and associated types that share a semantic. Turns out, a WGSL file is just that. This proposal introduces *implicit modules* as files.
+* global value and alias declarations are immutable and must be initialized. What if we could override them? if so, this is a simple mechanism for module specialization.
+  What if we also could override struct and functions declarations? In fact, all declarations?
+
+It proposes two extensions: [**Generic Declarations**](#generic-declarations) and [**Generic Modules**](#generic-modules) (+ optional [**Function-as-Type**](#function-as-type) [**Module-as-Type**](#module-as-type)).
```
mighdoll, line 17: I could write a github issue on the desire for generic declarations from my earlier work in stoneberry if that's still useful.

Generics.md
```diff
@@ -1,164 +1,386 @@
...
+}
+
+fn main() {
+  my_fn<u32, callback_implementor>();
+}
+```
+
+See [the map reduce example](#map-reduce).
+
+## Generic Modules
```
mighdoll, line 155: Is it clear that generic modules are the best way forward to add ergonmics to generic elements? If so, let's discuss here, if it's more debatable, perhaps we should split this part into a separate issue/PR.

Generics.md
```diff
@@ -1,164 +1,386 @@
...
+TL;DR instead, we could specify these constraints:
+* Unconstrained (or implicitly constrained by [AllTypes](https://www.w3.org/TR/WGSL/#alltypes-type).
+* Constrained by: a user-defined struct, `vec`, `mat`, `array`, `atomic`, `ptr`, `AbstractInt` or `AbstractFloat`: Must be specialized by a *constructible* subtype.
+* Constrained by a leaf type: `u32`, `i32`, `f32`, `f16`, `bool`: Must be specialized with a instance of that type (like the `N` parameter).
+
+## Generic Declarations
+
+We introduce a `@generic` attribute to define the generic parameters and constraints.
+```rs
+@generic(Num: AbstractFloat, T: MyStruct, N: u32)
```
mighdoll, line 100: Putting the generic paramaters between the fn name and the runtime parameters list is more traditional, no?
e.g. `fn foo<A: MyStruct, B>(a:A, b:B) { }`

Why not do that? Are you trying to minimize the new syntax? Does it create a parsing ambiguity?

Generics.md
mighdoll: in any case, I'll try to do a mock rewrite of one of the stoneberry files with the generic syntax.

Generics.md
```diff
@@ -1,164 +1,386 @@
...
+```
+
+### Using generic declarations
+
+I chose the template list notation, as it seems to be unambiguous in the grammar and is a familiar notation for generics. Besides, it is already used for type generators.
+
+```rs
+let res: array<f32, 10> = my_fn<f32, SomeStruct, 10>(3.5f, struct_inst);
+
+@generic(T: AbstractFloat)
```
mighdoll, line 120: AbstractFloat and AbstractInt are kinda wordy. How 'bout f64 and i64?

Generics.md
```diff
@@ -1,164 +1,386 @@
...
+After all, we are writing an extension, not a different language!
+It should look and feel just like the original stuff and extensions should be straight forward, self-explanatory even.
+
+## The proposal
+
+It builds on 3 observations
+
+* WGSL *already* has support for parametric types and functions, they are called [type-generators](https://www.w3.org/TR/WGSL/#type-generator) and [function overloads](https://www.w3.org/TR/WGSL/#builtin-functions). It even has some syntax for them.
+  The issue is, they are not allowed to be declared in user code. This proposal changes this.
+* A *module* is a collection of functions, structs and associated types that share a semantic. Turns out, a WGSL file is just that. This proposal introduces *implicit modules* as files.
```
mighdoll, line 13: The [glossary](https://github.com/wgsl-tooling-wg/wesl-spec/blob/main/GLOSSARY.md) defined Module as a single WESL file. Are implicit modules something different? 
(guessing it's just distinguishing vs the discussions of explicit namespaces within a file.)

Generics.md
```diff
@@ -1,164 +1,386 @@
...
+And we can add to that list:
+* `atomic<S>` is a subtype of `atomic<T>` if `S` is a subtype of `T`.
+* `ptr<S>` is a subtype of `ptr<T>` if `S` is a subtype of `T`.
+
+We can extend subtyping for user-defined types:
+* structure type `S1` is a subtype of structure type `S2` if, for all members `m` in `S2`: `S1.m` exists and `S1.m` is a subtype of `S2.m` (member covariance).
+This allows extending structs by narrowing down a member type, or by adding new members.
+* (optional) function signature `F1` is a subtype of function signature `F2` if,
+  * `F1` and `F2` have the same number of formal parameters,
+  * the `i`th formal parameter of `F1` is a subtype of the `i`th formal parameter of `F2` (parameter covariance),
```
mighdoll, line 41: Aren't function parameters parameters typically contravariant?

Generics.md
```diff
@@ -1,164 +1,386 @@
...
+  The issue is, they are not allowed to be declared in user code. This proposal changes this.
+* A *module* is a collection of functions, structs and associated types that share a semantic. Turns out, a WGSL file is just that. This proposal introduces *implicit modules* as files.
+* global value and alias declarations are immutable and must be initialized. What if we could override them? if so, this is a simple mechanism for module specialization.
+  What if we also could override struct and functions declarations? In fact, all declarations?
+
+It proposes two extensions: [**Generic Declarations**](#generic-declarations) and [**Generic Modules**](#generic-modules) (+ optional [**Function-as-Type**](#function-as-type) [**Module-as-Type**](#module-as-type)).
+They are orthogonal but they share the concepts of [*Type Hierarchy*](formalism-type-hierarchy) and [*Generic Type*](#formalism-generic-type) formalized below.
+We could support only one of the two and achieve the same expressivity AFAIK. But not the same ergonomics.
+
+## Formalism: Type Hierarchy
```
mighdoll, line 21: For most people (not all!) I think it'd read better with an example or two before we dig into the formalism.

Generics.md
```diff
@@ -1,164 +1,386 @@
...
+```
 
-    import binOpMax from ./ops;   // no generic variable specified yet
+## Implementation
 
-    export fn reduceMax<E>(elems: array<E, 2>) -> E { 
-      return binOpMax<E>(elems[0], elems[1]); // generic value applied at call site
-    }
-    ```
+Type checking does not have to be enforced by the linker, especially for runtime linking where we care about performance.
```
mighdoll, line 268: Nice feature!

Generics.md
```diff
@@ -1,164 +1,386 @@
...
+All of this proposal (except module-as-types) can be implemented with simple substitution.
+A compile-time linker or a IDE linter/langserver should implement type checking. Basically what Typescript does.
```
mighdoll, line 270: I think I know where you're going, but it's a little confusing because of course the TypeScript compiler does do type checking. (I think you mean that typical web development these days combines a fast transpiler that does no typechecking with a language server that does type checking)

Issue #14, opened 2024-09-02
Make wesl name official


stefnotch:
We should probably update our glossary and co to use the name "wesl".

k2d222:
I think we reached a consensus on that name, everyone seems to like it 😇

mighdoll:
yep, worth taking a pass through all the doc files.

k2d222:
I think we can close this.

PR #44, opened 2024-09-23
some doc cleanups, including partial fix for #14

mighdoll:


stefnotch (approved):
Thank you for doing this!

Generics.md
ncthbrt, line 1: Do you think we should maybe also remove this one for now?

k2d222 (approved):
Thanks for the cleanup 👍 
Minor comments:
* we should capitalize always WESL (instead of wesl).
* some indentation inconsistency in the code blocks
* some code blocks do not have `wgsl` after the code fence for syntax highlighting on gh.

Generics.md
mighdoll: regarding removing generics. I've thought that several times! My goal in this PR was elsewhere but I figured to at least adding some editorial for now as I drove by.

as far as removing generics section.. I've been tempted, because I think it's pretty obvious to us that we'll find something better than this generics draft and because generics isn't a critical feature for most of our customers. But I figured there's still some benefit in having a statement to new visitors that we're seriously considering generics as long as they're not confused that this draft is the plan.

mighdoll:
> Thanks for the cleanup 👍 Minor comments:

it's nice sometimes to have some low brain-wattage work I can do while traveling.

>     * we should capitalize always WESL (instead of wesl).
> 
>     * some indentation inconsistency in the code blocks
> 
>     * some code blocks do not have `wgsl` after the code fence for syntax highlighting on gh.

I will do another pass through to fix these and then merge.

PR #62, opened 2024-11-18
Update the importing spec

stefnotch:
This updates the importing spec

### DIFF
--- a/Imports.md
+++ b/Imports.md
@@ -86,6 +86,27 @@ import_collection:
 | '{' (import_path | item_import) (',' (import_path | item_import))* ','? '}'
 ```
 
+Equivalent LL(1) grammar for easier parsing
+```ebnf
+translation_unit:
+| import_statement* global_directive* global_decl* 
+
+import_statement:  
+| 'import' (import_relative | import_package) (import_collection | import_path_or_item) ';'  
+
+import_relative:
+| ('package' | 'super') '::' ('super' '::')*
+
+import_package:
+| ident '::'
+
+import_path_or_item:
+| ident ('::' (import_collection | import_path_or_item) | 'as' ident)?
+
+import_collection:
+| '{' (import_path_or_item) (',' (import_path_or_item))* ','? '}'
+```
+
 Where `translation_unit` and `ident` are defined in the WGSL grammar. 
 `ident`s are subject to the usual restrictions, meaning they cannot be keywords or reserved words.
 
--- a/Imports.md
+++ b/Imports.md
@@ -67,13 +67,10 @@ translation_unit:
 | import_statement* global_directive* global_decl* 
 
 import_statement:  
-| 'import' (import_relative | import_package) (import_collection | import_path_or_item) ';'  
+| 'import' import_relative? (import_collection | import_path_or_item) ';'  
 
 import_relative:
 | 'package' '::' | 'super' '::' ('super' '::')*
-
-import_package:
-| ident '::'
 
 import_path_or_item:
 | ident ('::' (import_collection | import_path_or_item) | 'as' ident)?
--- a/Imports.md
+++ b/Imports.md
@@ -109,6 +109,12 @@ To resolve the import, the recursive structure is flattened out. Then, one itera
        - (Re-exporting changes the path.)
        - (Inline modules do not have a path.)
 
+- `super`
+- `package`
+- `ident`
+
+
+
 For example
 
 ```wgsl
@@ -351,7 +357,6 @@ We encourage tooling authors to also implement source maps when implementing imp
 - Error Reporting. When Naga or Tint report an error in the generated WGSL code, we want to map the error location back to the WESL code.
 - Debugging. Eventually we hope to have a full toolchain of WESL ==generates a sourcemap==> WGSL ==generates a source map==> SPIR-V ==> RenderDoc shows WESL code
 
-
 ## Preprocessor
 
 How a preprocessor would interact with this proposal is an open question for a future proposal.
@@ -360,3 +365,7 @@ See [Conditional Compilation](./ConditionalCompilation.md).
 ## Scoped imports
 
 Allow imports that are only active within one function?
+
+## Importing `.wgsl` files
+
+Currently only `.wesl` files are importable. This restriction may be relaxed in the future.
--- a/Imports.md
+++ b/Imports.md
@@ -71,7 +71,7 @@ import_statement:
 | 'import' (import_relative | import_package) (import_collection | import_path | item_import) ';'  
 
 import_relative:
-| ('package' | 'super') '::' ('super' '::')*
+| 'package' '::' | 'super' '::' ('super' '::')*
 
 import_package:
 | ident '::'
@@ -95,7 +95,7 @@ import_statement:
 | 'import' (import_relative | import_package) (import_collection | import_path_or_item) ';'  
 
 import_relative:
-| ('package' | 'super') '::' ('super' '::')*
+| 'package' '::' | 'super' '::' ('super' '::')*
 
 import_package:
 | ident '::'
--- a/Imports.md
+++ b/Imports.md
@@ -287,7 +287,7 @@ Another drawback is that using the same name twice is impossible. In C-land, thi
 A future drawback is that "privacy" or "visibility" becomes very difficult to implement. Everything that is imported is automatically public and easily accessible.
 In C-land, the workaround is using header files. In other languages, such as Python, the convention ends up being "anything prefixed with an underscore `_` is private".
 
-## Typescript-like imports
+## TypeScript-like imports
 
 The Bevy team, with a large shader codebase, had a few wishes
 
@@ -335,7 +335,7 @@ automating shader composition for users.
 
 # Implementation
 
-This will be implemented in the [Javascript](https://github.com/wgsl-tooling-wg/wesl-js) and [Rust](https://github.com/wgsl-tooling-wg/wesl-rs) linkers.
+This will be implemented in the [JavaScript](https://github.com/wgsl-tooling-wg/wesl-js) and [Rust](https://github.com/wgsl-tooling-wg/wesl-rs) linkers.
 
 # Test cases
 
--- a/Imports.md
+++ b/Imports.md
@@ -61,7 +61,7 @@ A WESL program is composed of a tree of WESL modules.
 
 Imports must appear as the first items in a WESL file. They can import entire modules or individual  "importable items" (see [GLOSSARY.md](./GLOSSARY.md)).
 
-An import statement is parsed with the following LL(1) grammar, with spaces and comments allowed between tokens:
+An import statement is parsed with the following  grammar, with spaces and comments allowed between tokens:
 ```ebnf
 translation_unit:
 | import_statement* global_directive* global_decl* 
@@ -73,7 +73,8 @@ import_relative:
 | 'package' '::' | 'super' '::' ('super' '::')*
 
 import_path_or_item:
-| ident ('::' (import_collection | import_path_or_item) | 'as' ident)?
+| ident '::' (import_collection | import_path_or_item) 
+| ident ('as' ident)?
 
 import_collection:
 | '{' (import_path_or_item) (',' (import_path_or_item))* ','? '}'
--- a/Imports.md
+++ b/Imports.md
@@ -14,7 +14,7 @@ Finally, we want **multiple tools** which can compile WGSL-with-imports down to 
 
 ## Guide-level explanation
 
-The `import` statement extension brings items or entire modules into scope. To find the location of the imported items, it goes over the segments of the path, one by one.
+The `import` statement extension brings items or entire modules into scope. Import statements map to files with minimal searching.
 
 ```wgsl
 // Importing a single function using a relative path
@@ -38,8 +38,6 @@ fn main() {
 
 Both `bevy_ui` and `my` are packages in the current project. Language servers and related tools can look in a `wgsl.toml` file to find the location of the packages. This lets libraries be published to package managers, and users can import them with a simple syntax.
 
-The first part is the file path, which is assumed to have a `.wesl` file extension.
-
 Recursive import definitions are also supported, which leads to shorter import statements.
 
 ```wgsl
@@ -53,6 +51,13 @@ fn main() {
 
 }
 ```
+
+To find the relevant items, the following algorithm is used:
+
+Proceding left to right through the path segments, consider the segments `prev` and `seg`.
+1. if `prev.wesl` exists and includes WESL elements, check if `seg` is one of those elements, e.g. `fn seg` or `namespace seg`.
+1. else if the directory `prev/` exists, check to see if the file `seg.wesl` or the directory `seg/` is in the `prev/` directory;
+1. error if a `seg` is not found. 
 
 # Reference-level explanation
 
@@ -70,7 +75,7 @@ import_statement:
 | 'import' import_relative? import_path ';'  
 
 import_relative:
-| ('self' | 'crate' | 'super') '::' ('super' '::')* 
+| ('self' | 'crate' | 'super') '::' ('super' '::')*
 
 import_path:
 | (ident '::')+ (import_collection | item_import)  
@@ -100,7 +105,7 @@ To resolve the import, the recursive structure is flattened out. Then, one itera
 3. We repeatedly look at the next segment.
     1. Item in current module: Take that item. We must be at the last segment, otherwise it's an error.
     2. (Else if re-exported or inline module in current module: We continue with that module.)
-    3. Elso go to `current module path/ident.wesl`
+    3. Else go to `current module path/ident.wesl`
        - File found: We take that file as the current module.
        - File not found: We assume an empty module as the current module, and continue with that.
        - (Re-exporting changes the path.)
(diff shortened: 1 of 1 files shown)


ncthbrt:
Sorry. Didn't mean to mark as ready

stefnotch:
(No worries, I'll just update the title)

k2d222:
just a couple comments

the `crate::` keyword may not be the best choice, as crates are inherently a rust concept. I suggest `package` or `root` (preferred).

in rust 2018 edition, an import which starts with an ident (not `crate`, `self` or `super`) first looks at the child modules, then at external crates. Should we do that as well? The downside is that it costs a filesystem lookup and is ambiguous. The upside is that it simplifies `import self::foo::bar` to `import foo::bar`.

mighdoll:
> The `crate::` keyword may not be the best choice, as crates are inherently a rust concept. I suggest `package` or `root`.

I tried to summarize the then-current discussion in this [hack.md doc](https://hackmd.io/ljkByEcnQa2NdNLWed2M6Q#root-shortcut-for-import-statements
). There we used `package` and `super` (but not `self` or `crate`).

> first looks at the child modules, then at external crates. Should we do that as well? The downside is that it costs a filesystem lookup and is ambiguous. 

Simplifying a common import case seems goodness, though it's also tempting to defer until after M1 if it's nontrivial. I was thinking that most of the features in this [section](https://hackmd.io/ljkByEcnQa2NdNLWed2M6Q#Possible-Future-Features-in-the-Module-System) could be deferred til after our first rev.

stefnotch:
I updated the importing spec. @k2d222 I also changed the grammar a tiny bit, after realizing that I made a mistake and unintentionally disallowed `import super::foo`.

Imports.md
```diff
@@ -20,268 +14,230 @@ Finally, we want **multiple tools** which can compile WGSL-with-imports down to
...
 
-import_relative:  
-| ('.' | '..') '/' ('..' '/')*  
+import_statement:  
+| 'import' (import_relative | import_package) (import_collection | item_import | import_path) ';'  
 
-import_path:  
-| ident '/' (import_path | import_collection | item_import | star_import)  
+import_relative:
+| ('package' | 'super') '::' ('super' '::')*
```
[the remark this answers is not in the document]
stefnotch: Yes, I didn't find a strong enough reason to keep for M1.

Imports.md
```diff
@@ -20,268 +14,230 @@ Finally, we want **multiple tools** which can compile WGSL-with-imports down to
...
-The final module is produced by taking each imported module (in topological order, with the main module last), resolving and mangling all the globals, and joining the resulting pieces of code together.
+## Name Mangling
 
-All entry points and pipeline overridable constants from the imported modules are also mangled and land in the output.
+See [Name Mangling](./NameMangling.md)
 
-Dead code elimination is allowed, but not required.
+## Dead Code Elimination
+
+Linkers may choose to do dead code elimination, but it is a not-observable implementation detail.
```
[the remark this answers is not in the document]
stefnotch: Thanks for pointing this out! I added a note regarding them.

Imports.md
```diff
@@ -20,268 +14,230 @@ Finally, we want **multiple tools** which can compile WGSL-with-imports down to
...
-(In naga-oil entrypoints are lowered to normal functions.
-Not clear we should preserve that, it's against the spec,
-but noting current behaviour that is being used in the wild.)
+TODO: https://github.com/wgsl-tooling-wg/wesl-spec/issues/71
 
-## Preserved items
+## Entry points and pipeline overridable constants
 
-These items are preserved when importing a module. They are not imported, but will land in the final module with a mangled name.
+These items are preserved when importing a module. Their name must be preserved. They will land in the final module, if they are being referenced.
```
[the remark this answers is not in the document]
stefnotch: I changed it to use that wording 👍

stefnotch:
> one thing missing I think is describing where to look for the files:
> 
>     1. there is a root folder provided to the linker.

Should that be a root folder, or a root file? As in, should it be possible to write `package::Foo` to refer to a `Foo` struct that is defined in the WESL equivalent of a `lib.rs`?

>     2. how a relative path with self or super is converted to absolute path that starts at the root

I added this. Let me know if that's clear enough
"To get an absolute path to a module, one follows the algorithm above. In step 1, one takes the known absolute path of the `super` module, or the package.
The absolute path of the `super` module is always known, since the first loaded WESL file must always be the root module, and children are only discovered from there."



>     3. what happens if you have too many supers and exit the root?

Exiting the root should be an error


>     4. can we load both .wesl and .wgsl files? which takes priority?

I added a cop-out section at the end, which says "only wesl for now"

Imports.md
```diff
@@ -417,3 +362,7 @@ See [Conditional Compilation](./ConditionalCompilation.md).
 ## Scoped imports
 
 Allow imports that are only active within one function?
+
+## Importing `.wgsl` files
+
+Currently only `.wesl` files are importable. This restriction may be relaxed in the future.
```
mighdoll, line 368: Why not allow importing existing .wgsl files? Importing .wgsl makes it clear that WESL tooling is easy to adopt. 

If the only issue is duplicate named files like foo.wesl and foo.wgsl, we can pick a rule (perhaps drop the foo.wgsl, and linker warn)

Imports.md
```diff
@@ -370,44 +333,26 @@ For example, a linker can identify imports
...
-## Documentation comments
-
-This proposal works nicely with documentation comments in WGSL. This would allow library authors to document their shaders, which would be very useful for consumers.
-
 ## Source maps
 
-We encourage tooling authors to also implement source maps when implementing imports.
+We encourage tooling authors to also implement source maps when implementing imports. This aids
+- Error Reporting. When Naga or Tint report an error in the generated WGSL code, we want to map the error location back to the WESL code.
+- Debugging. Eventually we hope to have a full toolchain of WESL ==generates a sourcemap==> WGSL ==generates a source map==> SPIR-V ==> RenderDoc shows WESL code
```
mighdoll, line 355: (minor) the `sourcemap==>` syntax looks a little cramped. Maybe try spaces and -> or some alternative?

Imports.md
```diff
@@ -370,44 +333,26 @@ For example, a linker can identify imports
...
-This section is intended to encourage you as an author to think about the lessons from other languages, provide readers of your RFC with a fuller picture. If there is no prior art, that is fine - your ideas are interesting to us whether they are brand new or if it is an adaptation from other languages.
-
-Note that while precedent set by other languages is some motivation, it does not on its own motivate an RFC. Please also take into consideration that rust sometimes intentionally diverges from common language features.
--->
-
 # Implementation
 
-We are slowly working on adopting and adjusting the specification.
-This work is happening in multiple tools, including [a WGSL language server](https://github.com/dannymcgee/vscode-wgsl/), a [Typescript WGSL linker](https://github.com/mighdoll/wgsl-linker) and [Bevy's WGSL-imports library](https://github.com/bevyengine/naga_oil).
+This will be implemented in the [Javascript](https://github.com/wgsl-tooling-wg/wesl-js) and [Rust](https://github.com/wgsl-tooling-wg/wesl-rs) linkers.
```
mighdoll, line 338: JavaScript. (cap s)

Imports.md
stefnotch: I didn't want to add an additional "check for existence of different files" check to the linker rules. Node.js has that (js, cjs, mjs, ...), and from what I can gather, it was a mistake.

I would be envisioning some syntax level way of distinguishing between `import foo::bar` and `import foo::wgsl::bar`

stefnotch: But you're right, we should add a way of importing wgsl files as soon as possible.

k2d222: my implementation looks for the wesl file and if it does not exist it looks for a wgsl file and treats it as a wesl file. Idk if in the future we might want to treat vanilla wgsl files differently, e.g. disallow wesl extensions in them.

k2d222 (approved):
Thanks for your work @stefnotch ! I'm happy with the current state, I would approve merging after adressing @mighdoll 's comments.

Imports.md
stefnotch: Thanks 👍

Imports.md
stefnotch: I rephrased it as normal text

stefnotch (commented):
Make it clear what is in scope when I do `import foo::bar`, or when I do `import foo::{bar, baz as cat, foo}`.

Note that it's not necessary to parse dependent files when adding all those things to the module scope.

Imported items are in the module scope. They are not re-exported though.

Imports.md
stefnotch: TODO: Add a compatibility mode where wgsl files are allowed.
For publishing, they can simply be converted to wesl.

stefnotch:
A glob syntax in the `wesl.toml` could make sense for
- wgsl files are allowed/ignored
- this folder has tests, don't bother
- files can be scattered all over the project

k2d222:
Just thought of something. Currently the import path segments are WGSL identifiers. Maybe we want to restrict them more if they are supposed to match file/folder names. We want our packages to work on all platforms regardless of filesystem name restrictions.

This is two-fold: file names should be valid WGSL identifiers, and WESL module identifiers should be valid file names.

So for example the following should not be allowed (or just not reachable): 
* file `foo.bar.wesl` because `.` is not a valid identifier character. (same for all punctuation & symbols)
* file `foo bar.wesl` because it is two words.
* file `__foo.wesl` because WGSL idents starting with `_ _` are reserved.
* file `1foo.wesl` because WGSL idents cannot start with numbers.
* file `struct.wesl` because `struct` is a reserved WGSL word.
* perhaps folders `CON, PRN, AUX, NUL, COM1 - COM9, LPT1 - LPT9` because they are Windows reserved names?

Should we restrict to latin alphabet? Are there popular filesystems that don't support unicode letters? How do other languages deal with that issue?

k2d222:
> From what I can tell, other programming languages deal with this by ignoring the problem.

If that is true, then maybe we should do exactly the same :) Or just write a short note about it.

Imports.md
stefnotch: Adding this

stefnotch:
I added a new commit which addresses some review comments.

- Grammar change: Just swaps two elements. I'm thinking about including a version of the grammar that can be parsed with a single token lookahead, instead of two.
- Adds a note about where the root folder/root module is. (linker specific for now)
- Allows for `.wgsl` files
- Warns people of the terrors of filesystems

stefnotch:
Okay, I added a LL(1) grammar. @k2d222 or @mighdoll , could you double check if the grammar is the same?

Also, should I get rid of the non-LL(1) grammar? One downside of it is that it's more recursive, instead of using a `*` for repetition.

k2d222 (commented):
I vote for keeping only the LL1 to reduce noise.
Noticed that the grammar does not seem to allow `import foo;`, where foo is the root module of package foo.

Imports.md
```diff
@@ -20,70 +14,70 @@ Finally, we want **multiple tools** which can compile WGSL-with-imports down to
...
 
-import_relative:  
-| ('.' | '..') '/' ('..' '/')*  
+import_statement:  
+| 'import' (import_relative | import_package) (import_collection | import_path | item_import) ';'  
 
-import_path:  
-| ident '/' (import_path | import_collection | item_import | star_import)  
+import_relative:
+| ('package' | 'super') '::' ('super' '::')*
```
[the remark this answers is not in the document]
stefnotch: That's very smart, thank you!

stefnotch:
> Noticed that the grammar does not seem to allow `import foo;`, where foo is the root module of package foo.

Should it be allowed, and what should the semantics be? That would definitely be a module import, so inline it could only be used as `foo::bar`.
But then it's equivalent to `foo::bar` without an import statement?

k2d222:
@mighdoll my linker currently supports it so no additional effort on my side. I'm not going to argue if you think it's worth having  :+1: 

It might be useful for future `@publish` / re-exports too.

stefnotch:
Okay, I'm convinced, I'll tweak the grammar to allow that. Seems like it could be very useful in the future 👍

stefnotch:
Also, I'm allowing `import { foo, bar }`, because it simplifies the grammar, and Rust also allows it.

Imports.md
```diff
@@ -20,268 +14,243 @@ Finally, we want **multiple tools** which can compile WGSL-with-imports down to
...
 
 ```wgsl
 fn main() {
-    bevy_ui.quad(vec2f(0.0, 1.0), 1.0);
+    bevy_ui::quad(vec2f(0.0, 1.0), 1.0);
     let a = draw(3);
 }
 ```
 
 Both `bevy_ui` and `my` are packages in the current project. Language servers and related tools can look in a `wgsl.toml` file to find the location of the packages. This lets libraries be published to package managers, and users can import them with a simple syntax.
```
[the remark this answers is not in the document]
stefnotch: Ah, good catch! Fixed :)

Imports.md
```diff
@@ -20,268 +14,243 @@ Finally, we want **multiple tools** which can compile WGSL-with-imports down to
...
 
-star_import:
-| '*' ('as' ident)?
+import_relative:
+| 'package' '::' | 'super' '::' ('super' '::')*
 
-item_import:
-| ident ('as' ident)?
+import_path_or_item:
+| ident ('::' (import_collection | import_path_or_item) | 'as' ident)?
```
mighdoll, line 76: Perhaps something like this would be easier for the reader to parse? (I got a little crossed in the first read)
```ebnf
import_path_or_item:
| ident ('as' ident)?
| ident ('::' (import_collection | import_path_or_item))?
```

Imports.md
```diff
@@ -20,268 +14,243 @@ Finally, we want **multiple tools** which can compile WGSL-with-imports down to
...
-A star import behaves like a star import in Typescript, except that the name can be inferred from the file name. **Notably**, it does *not* import all the items individually. Instead, it groups them all under a single name!
 
 An item import imports a single item. The item can be renamed with the `as` keyword.
 
 An import collection imports multiple items, and allows for nested imports.
 
-## Examples
+To resolve the import, the recursive structure is flattened out. Then, one iterates over each segment, and looks it up one by one.
+
+1. We start with the first segment.
```
mighdoll, line 92: describe what flattened out means? I understand, but not sure the next reader would. Perhaps something about rewriting a path with collections to get rid of those and maybe ignoring the 'as' parts.

also note the iteration is left to right, e.g. iterations starts with the leftmost segment in the flattened import path.

Imports.md
```diff
@@ -20,268 +14,243 @@ Finally, we want **multiple tools** which can compile WGSL-with-imports down to
...
-```
+To resolve a module on a filesystem, one follows the algorithm above. 
+The root folder, or the root module, needs to be provided to the linker. This is currently a linker-specific API, and may change once we introduce a `wesl.toml`.
 
-</td>
-</tr>
-<tr>
-<td>
+Linkers are allowed to fall back to `.wgsl` files when a `.wesl` file cannot be found *in user code*.
+Published libraries will only contain `.wesl` files. Thus, linkers should rename `.wgsl` files to `.wesl` for publishing.
```
mighdoll, line 133: why does the suffix need to be translated? once we've ingested the file the linker looks it up by module path and ignores the suffix anyway.. 

If it doesn't matter, it's nice to expose the interop with wgsl, and have error messages that refer to the correct source file not a suffix translated one..

Imports.md
```diff
@@ -20,268 +14,243 @@ Finally, we want **multiple tools** which can compile WGSL-with-imports down to
...
+They will land in the final module, if they are statically accessed.
 
 - [Entry points](https://www.w3.org/TR/WGSL/#entry-points)
 - [Pipeline overridable constants](https://www.w3.org/TR/WGSL/#override-decls)
 
-## Identifier Resolution
+TODO: This will probably be improved after M1.
 
-The steps of identifier resolution are as follows:
+TODO: Do we have an issue that I can link to?
```
mighdoll, line 221: See #65

Imports.md
```diff
@@ -370,44 +340,26 @@ For example, a linker can identify imports
...
-This proposal works nicely with documentation comments in WGSL. This would allow library authors to document their shaders, which would be very useful for consumers.
-
 ## Source maps
 
-We encourage tooling authors to also implement source maps when implementing imports.
+We encourage tooling authors to also implement source maps when implementing imports. This aids
+- Error Reporting. When Naga or Tint report an error in the generated WGSL code, we want to map the error location back to the WESL code.
+- Debugging. Eventually we hope to have a full toolchain of WESL to WGSL to SPIR-V, with source maps at each step. In the end, it should be possible for RenderDoc to show the original WESL code.
 
 ## Preprocessor
```
mighdoll, line 364: conditional spec is in since this was originally written. are there any open questions wrt imports? If not, suggest editing this

Imports.md
```diff
@@ -20,268 +14,243 @@ Finally, we want **multiple tools** which can compile WGSL-with-imports down to
...
 
-The final module is produced by taking each imported module (in topological order, with the main module last), resolving and mangling all the globals, and joining the resulting pieces of code together.
+See [Name Mangling](./NameMangling.md)
 
-All entry points and pipeline overridable constants from the imported modules are also mangled and land in the output.
+## Dead Code Elimination
 
-Dead code elimination is allowed, but not required.
+Linkers may choose to do dead code elimination, but it is a not-observable implementation detail.   
+`const_assert` statements inside of functions need special treatment. Either those functions have to be kept,
```
mighdoll, line 252: Perhaps just say that we're going to drop dead functions and their `const_assert`s?
(Or make this a github issue instead of trying to explain our future options here.)

Imports.md
mighdoll, line 1: `lib.wesl` for packages? so packages can publish functions at the root level, as e.g. `random_wgsl` does. (or is there some other mechanism for that?)

Imports.md
stefnotch: Hmm, yes, that's easier to read. But it's no longer LL(1).
Then again, manually doing the conversion when implementing it is trivial.

I don't know what the preferred option is. @k2d222 any opinions?

stefnotch: Okay, the official recursive descent grammar also has enough cases where they opted for simplicity over LL(1)

Imports.md
stefnotch: Thank you!

Imports.md
stefnotch: I'll remove this bit, and if any questions ever arise, we can address them in the future 👍

Imports.md
stefnotch: Okay, I linked to the issue

Imports.md
stefnotch: The current mechanism is having a structure like
```
wesl.toml
shaders.wesl
shaders/
  foo.wesl
  bar.wesl
```

I'd say that this might just be good enough for M1, but we should revisit it. I also intentionally left that part of the spec under specified, so our linkers can try out a few different APIs

Imports.md
stefnotch: I expanded on this section a little bit. I now realize that I haven't really addressed how `as` aliases work.
I'm not entirely sure how to best describe them.

PR #72, opened 2025-01-31
dual license Apache2 + MIT

mighdoll:
This is adapted from the Rust terms. Please double check the edits, and/or peek at similar dual licensed projects.

Once we're good we can duplicate to all the projects. 

Also we'll need to do a relicensing step after this for previous contirbutors. (Bevy did relicensing with a github issue. Not sure what relicensing mechanism e.g. rust used.)


COPYRIGHT
stefnotch, line 1: Could we have the "Short version for non-lawyers:" heading? Just to make it explicit that the first part which says "license A and B" isn't the final source of truth. Instead the bottom part which says "license A or B" is what applies.

k2d222:
Looks good :+1:

COPYRIGHT
mighdoll: I'll remove that first bit, we can put the non-lawyer section in the README (as Rust also does). Most other projects I glanced at (e.g. bevy) don't seem to have this COPYRIGHT file at all, in favor of just the README and the two LICENSE files.

mighdoll: I updated the PR with non-lawyer license section in the README (like e.g. bevy)

mighdoll: see if that looks good and I'll merge.

stefnotch: This looks good 👍

PR #82, opened 2025-02-12
Require .wgsl fallback

stefnotch:
Is this good enough, or should we make the fallback mode optional?

Well technically, we can just say "fallback mode is mandatory" right now, and if we ever change our mind....*🎶 new edition time 🎵*


mighdoll:
LGTM

Issue #80, opened 2025-02-11
Incremental conditional-translation, shadowing and name mangling


k2d222:
With incremental conditional translation, the linker may generate an intermediate WESL code that still contains unresolved `@if` attributes. One code path could lead to shadowing a module-scope variable. If the module-scope variable name is mangled, we must rename all references to it. But with the unresolved conditional translation, we don't *know* what identifiers refer to it! 

problematic code:
```
const a = 8;

fn foo() {
   @if(cond)
   let a = 7;

   b = a; // does b refer to the module-scope `a` or the local declaration?
}
```

[relevant spec:](https://github.com/wgsl-tooling-wg/wesl-spec/blob/main/ConditionalTranslation.md#execution-of-the-conditional-translation-phase)
> NOTE: In case some features can only be resolved at runtime, a _WESL translator_ implementation can support feature specialization in two phases: A first pass is invoked with features determined at compile-time and returns a partially translated WESL source, replacing some of the _translate-time features_ with `true` or `false`. A second pass is invoked with features determined at run-runtime and returns a fully translated WGSL source.

**Solution**: we can rename local variables that bear the same name as the global variable they shadow. This situation is sufficiently obscure to deserve a clarification in the spec of [name mangling](https://github.com/wgsl-tooling-wg/wesl-spec/blob/main/NameMangling.md) later.

k2d222:
> We'll likely solve this by moving the name resolution and mangling to happen after conditional compilation.

It's suboptimal because it stops any preprocessing opportunities at built-time, the runtime linker has to do all the work. In particular we cannot bundle up files together which can be an issue if they are http-fetched.

> If there's any unresolved conditional compilation, we don't proceed further in the pipeline. This should be good enough for wesl-js.

It's good enough for M1 for sure 👍 let's see how this issue evolve when we get early adopters.

stefnotch:
Doing a binding structs transformation needs to happen after conditional compilation. Here is a case where we do not know if we have to rewrite the code in `do_particles_thing` or not

```rs
@if(FOO)
struct Bindings {
  @group(0) @binding(0) particles: ptr<storage, array<f32>, read_write>, 
}

@if(!FOO)
struct Bindings {
  particles: array<f32, 10>, 
}
    

fn do_particles_thing(b: Bindings) {
  let x = b.particles;
  let z = y.a;
}
```

mighdoll:
[...]

PR #85, opened 2025-02-18
Address some comments in ConditionalTranslation

k2d222:
Thanks @stefnotch for raising these issues. I added several examples and clarified the spec for: multiple `@if` attributes, order of attributes, and incremental translation.

There is still an unsolved issue: #80


stefnotch (approved):
Looks good

PR #101, opened 2025-03-01
Proposed Imports.md update

stefnotch:
This is one way of speccing out the inline usage and addressing my "what if a predeclared item gets added that conflicts with a package" concern.

Thinking forward: If WESL were to become a WGSL standard, then the WGSL authors would like to continue being able to add predeclared items. 
This makes sure that that remains the case. The alternative is that WGSL cannot add predeclared items, because they'd break existing code.


stefnotch:
TODO: Change this "- `ident` refers to an identifier that is in scope. If there is none. it falls back to a package." to only apply to inline usage.

PR #106, opened 2025-03-07
Fix links so that they work in both the public spec and the WESL website

JanMiksovsky:
Also fix a number of broken links with typos


stefnotch (approved):

stefnotch:
Thank you

PR #136, opened 2025-06-11
wesl.toml proposal

stefnotch:
Took me long enough, but here we are! A proposal for a `wesl.toml` format that should cover all cases, while remaining simple enough. Feedback most wanted

Fix https://github.com/wgsl-tooling-wg/wesl-spec/issues/5
Fix https://github.com/wgsl-tooling-wg/wesl-rs/issues/80
Fix https://github.com/wgsl-tooling-wg/wesl-spec/issues/57

### DIFF
--- a/WeslToml.md
+++ b/WeslToml.md
@@ -1,0 +1,87 @@
+# `wesl.toml` file format
+
+WESL shader projects have a standardized `wesl.toml` file, which describes the project setup. It is used by language servers and linkers[^1].
+
+The format looks as follows. As much of it is optional as possible.
+
+```toml
+[package] # Following the toml convention of grouping properties.
+
+# Optional, used for specifying the name within wesl shaders.
+# After all, the package manager's naming convention might clash with the WGSL naming conventions.
+# (E.g. camelCase vs snek_case)
+# Also deals with names like `@angular/animations`. Solves https://github.com/wgsl-tooling-wg/wesl-spec/issues/76
+name = "angular_animations" 
+
+# Not optional. Rust initially made it optional and it was a footgun.
+# It ended up being too easy to opt into the semantics of the oldest Rust edition.
+edition = "2025"
+
+# Description and version fields do not exist.
+# They match the package.json/Cargo.toml.
+# Semantics are "we are publishing a normal package that happens to also contain wesl code".
+# This lets us support wesl packages that come with host code!
+
+# Where are the shaders located. This is the path of `package::`.
+# We watch this directory for changes.
+root = "./shaders"
+
+# Some projects have large folders that we shouldn't react to. Globs for excluding them.
+exclude = [ "**/test" ]
+
+# Can be auto-inferred from the existence of a package.json.
+# Necessary when both a `package.json` and a `Cargo.toml` are present.
+# Inclusion of this field is gently encouraged.
+package-manager = "npm"
+
+# Optional dependencies section, automatically obtained from the package manager.
+[dependencies]
+ # Has no effect.
+foolib = {}
+ # Can be used for renaming packages. Now bevy in my code is called "cute_bevy".
+cute_bevy = { package = "bevy" }
+# Local imports, necessary for workspaces made of multiple separate packages.
+mylib = { path = "../mylib" }
+```
+
+
+## FAQ
+
+### Why restrict yourself to exactly one package manager in the `package-manager` field?
+
+It is unreasonable to require a language server to check both package managers for the same dependencies, and run everything twice.
+
+Tools are, however, free to support this sort of dual-publishing. They would have one primary package manager, and then attempt to mirror the structure for other package managers.
+
+### How do language servers find dependencies?
+
+Language server automatically obtain dependencies by asking the Typescript/Rust language servers.
+
+Alternatively, a language server can implement the whole resolution algorithm. This seems feasible for Javascript, and significantly harder for Cargo.
+
+In some cases, a language server can also obtain dependencies by calling the package manager's CLI.
+
+Each wesl implementation will need to work together with language server authors.
+
+### How does wesl find dependencies?
+
+Wesl implementations get the dependencies this by running a build script.
+
+In Rust, this means running a `build.rs` or a proc macro.
+
+In Javascript, there are multiple flavours of build scripts. 
+- Vite/rollup/... plugins
+- A Node.js script in the user's project, which ends up having the same semantics for dependency resolution
+
+
+## Defaults when `wesl.toml` is missing
+
+It is possible to have WESL shaders without a `wesl.toml`. In that case, we default to
+
+- All optional fields work as described above
+- `root` is undefined, therefore the `package::` syntax cannot be used. Only relative imports, like `super::foo` are supported
+- The edition field defaults to ... ?
+
+
+[^1]: Linkers are free to offer their own configuration, in addition to supporting the `wesl.toml` format.
+For example, wesl-js in the browser would be configured via a roughly equivalent JSON object. This lets wesl-js avoid bundling a whole toml file format parser.


WeslToml.md
```diff
@@ -0,0 +1,87 @@
...
+- A Node.js script in the user's project, which ends up having the same semantics for dependency resolution
+
+
+## Defaults when `wesl.toml` is missing
+
+It is possible to have WESL shaders without a `wesl.toml`. In that case, we default to
+
+- All optional fields work as described above
+- `root` is undefined, therefore the `package::` syntax cannot be used. Only relative imports, like `super::foo` are supported
+- The edition field defaults to ... ?
```
stefnotch: I am not sure what edition the language server should default to when no edition has been specified.

k2d222:
Curious about the dependency section. 
* Should all dependencies be listed there, or can it mix and match with those in Cargo.toml/package.json?
* Conversely, can it declare dependencies not listed in Cargo.toml/package.json?
* What's the point of the `foolib = {}`?

I suppose dependency version + unification is best left to the package manager. We should add a section for that in FAQ.

What about package metadata felds: version, description, license, authors... Should we not have them?

k2d222:
Thanks. The format is simple enough and seems to address the most pressing issues. I vote for 👍

WeslToml.md
```diff
@@ -0,0 +1,87 @@
+# `wesl.toml` file format
+
+WESL shader projects have a standardized `wesl.toml` file, which describes the project setup. It is used by language servers and linkers[^1].
```
mighdoll: and the packaging tools too, presumably..

WeslToml.md
```diff
@@ -0,0 +1,87 @@
...
+The format looks as follows. As much of it is optional as possible.
+
+```toml
+[package] # Following the toml convention of grouping properties.
+
+# Optional, used for specifying the name within wesl shaders.
+# After all, the package manager's naming convention might clash with the WGSL naming conventions.
+# (E.g. camelCase vs snek_case)
+# Also deals with names like `@angular/animations`. Solves https://github.com/wgsl-tooling-wg/wesl-spec/issues/76
+name = "angular_animations"
```
mighdoll: Hmm.. renaming on the publishing side seems like it would be too hard for library consumers to discover. I think renaming needs to happen on the consuming side. I like renaming at the import statement, but there are other options.

WeslToml.md
```diff
@@ -0,0 +1,87 @@
...
+# After all, the package manager's naming convention might clash with the WGSL naming conventions.
+# (E.g. camelCase vs snek_case)
+# Also deals with names like `@angular/animations`. Solves https://github.com/wgsl-tooling-wg/wesl-spec/issues/76
+name = "angular_animations" 
+
+# Not optional. Rust initially made it optional and it was a footgun.
+# It ended up being too easy to opt into the semantics of the oldest Rust edition.
+edition = "2025"
+
+# Description and version fields do not exist.
```
mighdoll: 👍

WeslToml.md
```diff
@@ -0,0 +1,87 @@
...
+
+# Optional, used for specifying the name within wesl shaders.
+# After all, the package manager's naming convention might clash with the WGSL naming conventions.
+# (E.g. camelCase vs snek_case)
+# Also deals with names like `@angular/animations`. Solves https://github.com/wgsl-tooling-wg/wesl-spec/issues/76
+name = "angular_animations" 
+
+# Not optional. Rust initially made it optional and it was a footgun.
+# It ended up being too easy to opt into the semantics of the oldest Rust edition.
+edition = "2025"
```
mighdoll: agreed. `edition` should be required for published libraries. Could we get away with `edition` optional for apps? It would be nice if `wesl.toml` could be entirely optional for simple, typical apps.

WeslToml.md
```diff
@@ -0,0 +1,87 @@
...
+
+# Description and version fields do not exist.
+# They match the package.json/Cargo.toml.
```
mighdoll: Ought description and version fields exist at all in the bundled shaders in js or rs or is it sufficient to rely on native packagers?

WeslToml.md
```diff
@@ -0,0 +1,87 @@
...
+# Semantics are "we are publishing a normal package that happens to also contain wesl code".
+# This lets us support wesl packages that come with host code!
```
mighdoll: +1, agreed.

also can our std directory structure support host code in both rust and js? I think/hope so.

WeslToml.md
```diff
@@ -0,0 +1,87 @@
...
+
+# Where are the shaders located. This is the path of `package::`.
+# We watch this directory for changes.
+root = "./shaders"
+
+# Some projects have large folders that we shouldn't react to. Globs for excluding them.
+exclude = [ "**/test" ]
```
mighdoll: I think we probably need `include` too. Technically, one of `include` or `exclude` is sufficient. But in practice I think particular projects will think about it one way or another, and we need flexibility to support integration into various projects. Lygia would nearly have broken recently with only `exclude`, because a PR with a new directory under shaders was added by a non-wesl-savvy contributor.

WeslToml.md
```diff
@@ -0,0 +1,87 @@
...
+
+# Can be auto-inferred from the existence of a package.json.
+# Necessary when both a `package.json` and a `Cargo.toml` are present.
+# Inclusion of this field is gently encouraged.
```
mighdoll: why encourage this field even gently for a typical rust or js app?

WeslToml.md
```diff
@@ -0,0 +1,87 @@
...
+```
+
+
+## FAQ
+
+### Why restrict yourself to exactly one package manager in the `package-manager` field?
+
+It is unreasonable to require a language server to check both package managers for the same dependencies, and run everything twice.
+
+Tools are, however, free to support this sort of dual-publishing. They would have one primary package manager, and then attempt to mirror the structure for other package managers.
```
mighdoll: It's an interesting case! If you're dual publishing, is this best controlled in wesl.toml or in a language server config?

WeslToml.md
mighdoll: > I am not sure what edition the language server should default to when no edition has been specified.

Perhaps the language server could peek at package.json / cargo.toml? There won't be that many stable versions of wesl and we'd know the semver where e.g. stable_2026 begins.

WeslToml.md
```diff
@@ -0,0 +1,87 @@
...
+
+Language server automatically obtain dependencies by asking the Typescript/Rust language servers.
+
+Alternatively, a language server can implement the whole resolution algorithm. This seems feasible for Javascript, and significantly harder for Cargo.
+
+In some cases, a language server can also obtain dependencies by calling the package manager's CLI.
+
+Each wesl implementation will need to work together with language server authors.
+
+### How does wesl find dependencies?
```
mighdoll: Since [#170](https://github.com/wgsl-tooling-wg/wesl-js/issues/170), the wesl-js tools (cli, plugins, packager) will automatically find library dependencies w/o requiring the user to mention them explicitly in `wesl.toml`. It's a much nicer experience than the M1 requirement that the user maintain duplicate lists of dependencies (once in `package.json` and again in `wesl.toml`). 

Can we extend that approach to wesl-rs too?

WeslToml.md
BenjaminBrienen: It would have to be the oldest edition. This also makes the earlier comment not really true. Why do we want to support letting authors not writing wesl.toml?

WeslToml.md
```diff
@@ -0,0 +1,87 @@
...
+# We watch this directory for changes.
+root = "./shaders"
+
+# Some projects have large folders that we shouldn't react to. Globs for excluding them.
+exclude = [ "**/test" ]
+
+# Can be auto-inferred from the existence of a package.json.
+# Necessary when both a `package.json` and a `Cargo.toml` are present.
+# Inclusion of this field is gently encouraged.
+package-manager = "npm"
```
stefnotch: Add a list of currently accepted values for this field

WeslToml.md
stefnotch: I think the previous rationale was "it's annoying to always have to include a config file".
But I suspect that there's no way around requiring a minimal config file.

WeslToml.md
```diff
@@ -0,0 +1,87 @@
+# `wesl.toml` file format
+
+WESL shader projects have a standardized `wesl.toml` file, which describes the project setup. It is used by language servers and linkers[^1].
+
+The format looks as follows. As much of it is optional as possible.
+
+```toml
+[package] # Following the toml convention of grouping properties.
+
+# Optional, used for specifying the name within wesl shaders.
```
stefnotch: Change this phrasing to explicitly mention wesl import statements.

WeslToml.md
BenjaminBrienen: Adding support for "implied wesl.toml" is easier than removing support.

WeslToml.md
stefnotch: Make it clear that we are not committing to an edition yet. How does `unstable_2025` sound?

BenjaminBrienen: Good

WeslToml.md
mighdoll: It's nice to have the config file optional for apps (not for libraries), if we can manage it. I think it can be optional.

mighdoll: I've been writing some docs and examples in recent weeks, and it's nice to make the steps for adding wesl to a project as minimal as possible to reduce adoption friction. 

Here's one example that'll eventually make it into lygia docs. The instructions will say install and then make two changes:

https://github.com/mighdoll/lygia-static-example/blob/d5551269735f492caac62c4456b317fc35b2cfab/shaders/app.wesl#L2
https://github.com/mighdoll/lygia-static-example/blob/d5551269735f492caac62c4456b317fc35b2cfab/src/main.ts#L2

WeslToml.md
```diff
@@ -0,0 +1,87 @@
...
+# Necessary when both a `package.json` and a `Cargo.toml` are present.
+# Inclusion of this field is gently encouraged.
+package-manager = "npm"
+
+# Optional dependencies section, automatically obtained from the package manager.
+[dependencies]
+ # Has no effect.
+foolib = {}
+ # Can be used for renaming packages. Now bevy in my code is called "cute_bevy".
+cute_bevy = { package = "bevy" }
```
stefnotch, line 31: Make it clear what names go here when a package has a different name for wesl.

(Left side: wesl name. Right side: Original name.)

BenjaminBrienen:
I think dependencies should be explicitly included to make it easier for language servers and other tools to discover relevant files, make it more clear to developers what they are working with, and so that packages that happen to include WESL code, but are undesirable to include, can be left out.

Some additional reasons that may or may not be valid:
- The package is used for utilities or bindings, but its shaders are unused.
- The shaders in the package are intended for a different platform/target architecture.
- The shaders in the package are included for testing or development purposes of that dependency only.

stefnotch:
This seems quite reasonable. I'm happy to make the dependency list explicit.

In a world with an explicit dependency list, we still have a lot of strategies for making it nice and easy.
e.g. `wesl init` which would automatically create a wesl.toml file with the desired dependencies.
e.g. The error message for `import unknownlib::foo` can suggest adding the library to the wesl toml. 
etc

@k2d222 @mighdoll What are your opinions?

BenjaminBrienen:
I disagree that it makes the user's experience worse. I think it is more clear for developers to see a list of their WESL project's dependencies in the WESL project file, especially if they don't actually want everything possible to be imported. I'm open to experimenting with it either way.

mighdoll:
> I disagree that it makes the user's experience worse. 

Isn't it worse in the typical case where adding a dependency means remembering to add it in two places, and debugging if you mispell, remove a dependency, etc.? I certainly made that mistake many times in writing the first examples..

I totally agree that required dependencies in wesl.toml is workable and will be aok if we need to go that way. I'm not certain, but so far I think required dependencies are usually worse for the user so I'm arguing the other way so we get to the root of the issues involved.

> I think it is more clear for developers to see a list of their WESL project's dependencies in the WESL project file, especially if they don't actually want everything possible to be imported. 

That's an interesting scenario.. so the scenario is that you've installed a shader package with npm or cargo, but want to make sure it's inaccessible from wesl? what are you thinking about? maybe a shader library that's useful in tests but you don't want to accidentally include in the app or library?

BenjaminBrienen:
That's possible, yeah. Maybe my game uses bevy for everything except the shaders and I don't want to import bevy's shaders in my WESL project. If adding the dependency in 2 places is a common tripping point, then we could make that easier by having `wesl add xpackage` call `cargo add xpackage` *and* add it in wesl.toml. If they call `cargo add xpackage` themselves, maybe we shouldn't assume that whatever shaders happen to ship with that package are intended to be added to the wesl project.

stefnotch:
I was thinking about the various different runtimes and scenarios (like "Node.js without a bundler", "Deno but it's compiled into a single script", ...)

And I think I have one that greatly benefits from an explicit up-front list of dependencies:
*Rust with shader hot reloading*

So, when Bevy, or a Bevy game is being developed, the following happens
1. Rust compiler
2. WESL-rs build script. Resolves all libraries. (Roughly does what the wesl-js build plugin also does.)
3. Now we have an executable.
4. The executable is started.
5. Since recompiling Rust is slow, we register a file watcher. It watches the wesl files. 
6. Wesl file is changed, we run WESL-rs at runtime to translate the shader.

However, at runtime, we *cannot* resolve new dependencies. Which already makes a hello world example fail in a horrible way. The moment one adds their first import from a not-yet-imported Bevy dependency is when WESL-rs would throw an error.

stefnotch:
Yes, restarting for recompiling is unavailable when changing the Cargo.toml. I, as a Rust user, would expect that.


What I would not expect is the restarting when using a new shader library. That would permanently look like a bug to me. It is a case of "thinking like a compiler, instead of thinking like a user". We'd have to double check with the Bevy folks, but I assume that they'd also view this as an issue.

With that in mind, for WESL-rs, I believe I would prefer an explicit list being mandatory. 
(Unless there's a smart workaround that I'm overlooking.)

WeslToml.md
stefnotch: The name field will not be a thing in favour of automatic name translations and users using explicit names

WeslToml.md
```diff
@@ -0,0 +1,113 @@
+# `wesl.toml` file format
+
+WESL shader projects have a standardized `wesl.toml` file, which describes the project setup. It is used by language servers and linkers[^1].
+
+The format looks as follows. As much of it is optional as possible.
+
+```toml
+[package]
+# Which wesl edition are we using
```
mighdoll: rather than using 'edition' to explain the 'edition' field, say 'version of WESL'. e.g. 'version of WESL used in the local shaders'.

WeslToml.md
```diff
@@ -0,0 +1,113 @@
...
+edition = "unstable_2025"
+
+# Description and version fields do not exist.
+# They match the package.json/Cargo.toml.
+# Semantics are "we are publishing a normal package that happens to also contain wesl code".
+# This lets us support wesl packages that come with host code!
+
```
mighdoll: Move to discussion, confusing for the less initiated to mention fields that don't exist.

WeslToml.md
```diff
@@ -0,0 +1,113 @@
...
+# Where are the shaders located. This is the path of `package::`.
+# We watch this directory for changes.
+root = "./shaders"
```
mighdoll, line 17: can we make this the default? That'll keep wesl.toml entirely optional for most getting started apps.

WeslToml.md
```diff
@@ -0,0 +1,113 @@
...
+
+# Optional. Default value is all wesl and wgsl files in the root directory.
+include = [ "shaders/**/*.wesl", "shaders/**/*.wgsl" ]
```
mighdoll, line 20: mention that this is glob format. (presumably?) I think the version you have is more clear than `*.w[eg]sl`.

also, mention somewhere that paths are relative the directory containing the wesl.toml file.

WeslToml.md
```diff
@@ -0,0 +1,113 @@
...
+
+# Optional.
+# Some projects have large folders that we shouldn't react to. 
+exclude = [ "**/test" ]
+
```
mighdoll, line 25: clarify default value.

WeslToml.md
```diff
@@ -0,0 +1,113 @@
...
+# Inclusion of this field is encouraged.
+package-manager = "npm"
+
+# Lists all used dependencies
+[dependencies]
+# Shorthand for `foolib = { package = "foolib" }`
+foolib = {}
+ # Can be used for renaming packages. Now bevy in my code is called "cute_bevy".
+cute_bevy = { package = "bevy" }
+# File path to a folder with a wesl.toml. Simplest kind of dependency.
```
mighdoll, line 32: does this work too? `@angular/shaders = { package = "ang_shaders"}`?

I like the idea of client renaming of packages at the very least to handle the cases where auto-renaming is insufficient. For a 'hybrid' library, is it better to rename in the package manager and not in the wesl.toml? I suppose things like reflection or reified references will make the wesl module paths visible in host code.. I'm not sure this matters, just raising the question so you can help think it through.

WeslToml.md
```diff
@@ -0,0 +1,113 @@
...
+mylib = { path = "../mylib" }
+```
+
+
+[^1]: Linkers are free to offer their own configuration, in addition to supporting the `wesl.toml` format.
```
mighdoll: I think the key point here is that wesl.toml is not used at runtime, it's for command line / ide tools. 
At one point I thought we'd translate to json as this footnote suggests, but now the [WeslBundle](https://wesl-lang.dev/docs/Publishing-Packages#npm-shader-package-format) is more of a second cousin than a twin to wesl.toml.

WeslToml.md
```diff
@@ -0,0 +1,113 @@
...
+For example, wesl-js in the browser would be configured via a roughly equivalent JSON object. This lets wesl-js avoid bundling a whole toml file format parser.
+
+
+## Semantics
+
+### `edition` field
+
+Supports all wesl editions. Currently only `unstable_2025` is accepted, since we do not have an edition yet.
```
mighdoll: 'unstable_2025' is a sort of edition.. It's the contents of the edition field. (and we might someday have unstable_2026, which wouldn't be precise but we'd use to tell tools things were fresher than 'unstable_2025')

Perhaps something like:
Currently only `unstable_2025` is accepted for the `edition` field. (No incompatible changes are planned for WESL, but we have not yet committed to stable version for long term support.)

WeslToml.md
```diff
@@ -0,0 +1,113 @@
...
+All used dependencies are explicitly listed here.
+
+### `dependencies = "auto"`
+
+Instead of specifying a list of dependencies, one can specify `dependencies = "auto"`.
+Then, the available dependencies are automatically inferred from the list of installed libraries. 
+
+This is an ecosystem specific feature. It is supported by
+
+- npm: enabled by default. Exact semantics will be figured out.
```
mighdoll: I think it's something like this: packages are resolved from WESL module paths using the standard node package resolution algorithm for esm modules, with double colons in WESL converted into slashes for resolution. As with normal package resolution, npm packages with `export` sections may provide subpath exports. So WESL `mypkg::foo::bar::baz()` will use the `"./foo":` export in `mypkg`'s `package.json` if it exists. (I'm presuming that auto-renaming for '@' and '-' would not supported by `dependencies = "auto"`)

WeslToml.md
```diff
@@ -0,0 +1,113 @@
...
+
+
+## Q&A
+
+### How do language servers find dependencies?
+
+Language server automatically obtain dependencies by asking the Typescript/Rust language servers.
```
mighdoll: Is this still true? (I'd presume it'd be easier to just call a library for npm, and I though there was some discussion of a cargo library on list)

WeslToml.md
```diff
@@ -0,0 +1,113 @@
...
+
+Each wesl implementation will need to work together with language server authors.
+
+### How does wesl resolve dependencies?
+
+Wesl implementations resolve the dependencies this by running a build script.
+
+In Rust, this means running a `build.rs` or a proc macro.
+
+In Javascript, there are multiple flavours of build scripts.
```
mighdoll: maybe just reference: https://wesl-lang.dev/docs/JavaScript-Builds

WeslToml.md
```diff
@@ -0,0 +1,113 @@
...
+
+Alternatively, a language server can implement the whole resolution algorithm. This seems feasible for Javascript, and significantly harder for Cargo.
+
+In some cases, a language server can also obtain dependencies by calling the package manager's CLI.
+
+Each wesl implementation will need to work together with language server authors.
+
+### How does wesl resolve dependencies?
+
+Wesl implementations resolve the dependencies this by running a build script.
```
mighdoll: 'this' looks extraneous

do you want to explain what 'resolving dependencies' means?

WeslToml.md
```diff
@@ -0,0 +1,113 @@
...
+
+
+## Defaults when `wesl.toml` is missing
+
+**Creating a `wesl.toml` is recommended.**
+
+However, we support having WESL shaders without a `wesl.toml`. In that case, we default to
+
+- All optional fields work as described above
+- `root` is undefined, therefore the `package::` syntax cannot be used. Only relative imports, like `super::foo` are supported
```
mighdoll: If we're going to default to having `include = [shaders/**/*.w[eg]sl]` seems like we should default to having `root = "shaders"`, and then `package::` works fine.

WeslToml.md
```diff
@@ -0,0 +1,113 @@
...
+- `edition` defaults to the latest wesl edition that *the tool* knows of.
+
+Not specifying an edition comes with the explicit risk of *tools no longer understanding your wesl code after an update*.
```
mighdoll: Perhaps mention that `edition` is required in every `wesl.toml`. 

Suggest making this closing note positive rather than scary.
Not having a wesl.toml is fine for getting started with small projects, but adding a wesl.toml is recommended so that future tool versions can reliably recognize older versions of wesl.

mighdoll (changes requested):

WeslToml.md
stefnotch: I think it's sufficient to rely on native packagers.

We can always add the fields later if it turns out that they would have been useful.

WeslToml.md
stefnotch: I don't see an immediate reason why it wouldn't be possible.

Separate issue :)

WeslToml.md
stefnotch: Good idea 👍

WeslToml.md
stefnotch: We could, but I'd prefer this to happen in a separate change. I haven't thought through the consequences of that.

WeslToml.md
stefnotch: The other way around would be the option that works.
`ang_shaders = { package = "@angular/shaders" }`

Though `_shaders` is that naming convention where one puts the type in the name. One should just do
`angular = { package = "@angular/shaders" }`

WeslToml.md
stefnotch: Any suggestions on how to phrase that?

WeslToml.md
stefnotch: Sounds good, added some of that phrasing

WeslToml.md
stefnotch: I'm removing that section. We can add it again once we've actually figured out and implemented it.

WeslToml.md
stefnotch: Suggestions on alternative phrasing?

WeslToml.md
stefnotch: Oh hmm, I haven't thought about that. I don't know what the default for the `include` field would be, unless we also specify the default for the `root` field.

WeslToml.md
BenjaminBrienen: I agree. I think renaming it in the consumer's wesl.toml like `angular_animations = { package = "@angular/animations" }` is cleaner.

stefnotch: Sounds good. The `name` field got removed, since everyone I asked seems to be of the same opinion.

WeslToml.md
stefnotch: The default could be "folder of package.json/Cargo.toml" with "./shaders"

WeslToml.md
stefnotch: We could default to `root` starting at the closest package manager file. As in, a `package.json`, or a `deno.json` or a `Cargo.toml` or a `rust-project.toml`

The problems start once we expand into ecosystems that don't have such a file. Like *vanilla JS* or C++.
Then every wesl tool would search for the file, going all the way to the root directory, and fail.

WeslToml.md
```diff
@@ -0,0 +1,118 @@
...
+[package]
+# Version of WESL used in the local shaders
+edition = "unstable_2025"
+
+# Where are the shaders located. This is the path of `package::`.
+# We watch this directory for changes.
+root = "./shaders"
+
+# Optional. Default value is all wesl and wgsl files in the root directory.
+include = [ "shaders/**/*.wesl", "shaders/**/*.wgsl" ]
```
stefnotch, line 20: Add some documentation akin to https://www.typescriptlang.org/tsconfig/#include

That would make it clear that "absolute paths" and other such weird cases aren't supported here.

WeslToml.md
mighdoll: perhaps we just check/assume the current directory when running a command line tool, w/o searching up the tree for a file.

stefnotch: I think the language server should not assume that the current directory is the right one.

WeslToml.md
```diff
@@ -0,0 +1,113 @@
+# `wesl.toml` file format
+
+WESL shader projects have a standardized `wesl.toml` file, which describes the project setup. It is used by language servers and linkers[^1].
+
+The format looks as follows. As much of it is optional as possible.
```
mighdoll, line 5: Simple applications can rely on defaults and not use a `wesl.toml` file, but all publishable shader libraries and many applications will need a `wesl.toml` file to customize settings.

The format looks as follows. All fields are optional except `edition`.

WeslToml.md
mighdoll: suggest just stopping after:
Currently only unstable_2025 is accepted for the edition field.

WeslToml.md
mighdoll: So I think `package::` syntax is always usable... but did we discuss which mechanism for finding the root if unset? I recall we discussed ecosystem specific cargo.toml and package.json, but not sure if that applied to this situation.

WeslToml.md
```diff
@@ -0,0 +1,113 @@
...
+In Javascript, there are multiple flavours of build scripts. 
+- Vite/rollup/... plugins
+- A Node.js script in the user's project, which ends up having the same semantics for dependency resolution
+
+
+## Defaults when `wesl.toml` is missing
+
+**Creating a `wesl.toml` is recommended.**
+
+However, we support having WESL shaders without a `wesl.toml`. In that case, we default to
```
mighdoll: Simple applications can get started without a `wesl.toml`. In that case...

WeslToml.md
mighdoll: - `edition` defaults to the latest wesl edition that the tool knows of. (Best practice is to include a `wesl.toml` file and explicitly control the edition for stability when later upgrading tools.)

WeslToml.md
mighdoll: maybe just remove the footnote on the argument that the details of whether wesl-js needs a toml parser are interesting to us, but not interesting for everyone else.

WeslToml.md
stefnotch: It's much easier to read a single field than it is to keep up with an ever increasing list of files to check. 

Different tools will have different lists, so they can end up looking at the same wesl project and assume different semantics.

WeslToml.md
stefnotch: I don't actually know, I haven't tried it out yet.

BenjaminBrienen:
It isn't clear why adding a wesl.toml makes builds more "consistent".

WeslToml.md
stefnotch: Done, but that uncovered a new question.

WeslToml.md
```diff
@@ -0,0 +1,135 @@
...
+
+The default value of `include` is all wesl and wgsl files in the `root` directory, recursively.
+The default value of `exclude` is an empty array.
+
+The patterns are glob patterns, which support
+- `*` for zero or more characters in file/folder names
+- `?` for one character in a file/folder name
+- `**/` for any directory nested to any level
+
+The globs must match file names, and the last part of the patterns must contain a file extension.
```
stefnotch: Wait, how should the globs work? Should we allow including folders  `include = ["shaders"]` , instead of writing  `include = ["shaders/**/*.wesl"]`?

If yes, what are the semantics? Does that include `.wgsl` files or does it only include `.wesl` files?

WeslToml.md
stefnotch: I'm just removing this section, since I think it's no longer useful

WeslToml.md
```diff
@@ -0,0 +1,123 @@
...
+exclude = [ "**/test" ]
+
+# Lists all used dependencies
+[dependencies]
+# Shorthand for `foolib = { package = "foolib" }`
+foolib = {}
+ # Can be used for renaming packages. Now bevy in my code is called "cute_bevy".
+cute_bevy = { package = "bevy" }
+# File path to a folder with a wesl.toml. Simplest kind of dependency.
+mylib = { path = "../mylib" }
```
BenjaminBrienen, line 33: The meaning of "path" here is not actually specified.

WeslToml.md
```diff
@@ -0,0 +1,123 @@
+# `wesl.toml` file format
+
+WESL shader projects have a standardized `wesl.toml` file, which describes the project setup. It is used by tools such as language servers and linkers[^1].
+
+The format looks as follows. As much of it is optional as possible.
+
+```toml
+[package]
+# Version of WESL used in the local shaders
```
BenjaminBrienen:
```suggestion
# Version of WESL used in this project.
```

"local" implies to me that local path dependencies will be forced to use this edition.

WeslToml.md
```diff
@@ -0,0 +1,123 @@
...
+
+### `edition` field
+
+Specifies which edition is used by wesl. All wesl editions can be used.
+
+- Mandatory
+  - Falls back to latest edition that tools know.
+- Currently only `unstable_2025` is accepted, since we do not have an edition yet.
+  - No incompatible changes are planned, but we have not committed to stable version for long term support.
+
```
BenjaminBrienen, line 54:
```suggestion
- Dependencies are interpreted using the edition that they declared, not the edition of the consuming package.

```

Issue #132, opened 2025-05-17
overridable library constants in WESL


mighdoll:
WESL/WGSL doesn't currently have a way to define overridable constants in libraries and then override them in shader code. The constant names should be probably be 'manageable' in the same sense as #131.

The Lygia library often uses `#define` constants to allow users to parameterize shader code from within the shader, e.g. [here](https://github.com/patriciogonzalezvivo/lygia_examples/blob/e58928c8060fdc557da37a4d288fb5e65d77faed/lighting_atmosphere.frag#L27C1-L33C1):

```c
#define ATMOSPHERE_SUN_POWER 20.0
#define ATMOSPHERE_LIGHT_SAMPLES 8
#define ATMOSPHERE_SAMPLES 26
#define ATMOSPHERE_STARS_LAYERS 3
#define ATMOSPHERE_STARS_ELEVATION u_time * 0.01
#define ATMOSPHERE_STARS_AZIMUTH u_time * 0.003
#include "lygia/lighting/atmosphere.glsl"
```

The constants are then used not just in the current module but also in `#include`d modules. 

cc: @patriciogonzalezvivo

Issue #131, opened 2025-05-16
Packaging Conditions in Libraries


mighdoll:
Let's say a package `mylib` uses `@if(MOBILE)`. I'd argue that conditions like MOBILE should have the following properties:
1. host app code can set MOBILE
2. build tools can set MOBILE
3. importers (app or library wesl that imports mylib) can set MOBILE
4. importers can rename MOBILE to avoid name conflicts with other libs
5. importers can decide whether MOBILE is visible to consuming packages or host code

WESL M1 supports the first two properties. I'd argue that the last three are important for modularity (package effects should be containable) and orthogonality (shader control should be available from both host and shader code).

Lygia for example uses conditions like `@if(YUV_SDTV)` and `@if(CIE_D50)` to control color space conversion, presumably because any given app typically only wants one mode for the relevant conversion.

k2d222:
While I agree on principle, the dependency tree can make it complex to resolve conditions if say packages A and B depend on a package C with different features.

If we want to avoid code duplication, the dependency resolution algorithm should unify features. (see rust doc https://doc.rust-lang.org/cargo/reference/features.html#feature-unification)
This has implications on the design and use of features, such as:
* features should be additive
* avoid exclusive features
* all combinations of features should be valid

A more complex algorithm could instead duplicate code only when it depends on conflicting features. This is similar to how packages with incompatible versions are duplicated.
Note that this solution introduces side-effects, e.g. duplicating a package duplicates its `var<private>` variables.

point 4. assumes that features are transparently transmitted to dependent packages. In rust, a package can only turn ON features in its dependencies, and feature names are scoped within a package. 
Though, the story is different with the C preprocessor where features have de facto a 'compilation unit' scope.

for `wesl-rs` it would be tempting to export wesl features as Cargo features, e.g. MOBILE would be accessible as wesl-mobile in the toml. This way we let cargo do the heavy lifting but we loose control over the behavior.

mighdoll:
Good points! 

I agree, the orig point 4 implies too much. The key is that there's away to arrange that condition names from packages _avoid conflict_. Having a global namespace for condition names with a renaming facility is one way to do that. But having condition names be scoped to packages would address the that concern nicely too.

condition unification for shared dependencies sounds fancy. 
- If it's not feasible, duplicating libraries when necessary seems fine. Just like we'd duplicate on different library versions. 
[...]

PR #146, opened 2025-08-25
Module Parameters Proposal

k2d222:
Rough draft

See TODOs for questions to resolve. Also, maybe we should do just consts to get started.


ModuleParameters.md
```diff
@@ -0,0 +1,100 @@
...
+
+Module parameters defined or published (TODO: publish spec) in the Root Module are called Shader Parameters. Shader Parameters can be set from the host code with a linker-specific API.
+
+NOTE: root module declarations are not reachable from other modules (due to cyclic imports) so shader parameters can only be replaced from the host code. (TODO: should we allow import cycles?)
+
+### Conditional Translation
+
+Changes to the Conditional Translation spec:
+
+* Module parameters of type `bool` can be used as feature flags for conditional translation. (TODO: integer conditionals)
```
mighdoll, line 92: Integer conditionals sounds good, but suggest we just note it as a future work item for this PR.

ModuleParameters.md
```diff
@@ -0,0 +1,100 @@
...
+
+Modules containing parameter declarations are called generic modules. Each usage of a generic module, with different parameters, forms a concretization of the module. Declarations inside different concretizations of the same module are duplicated. (TODO: what happens with declarations that don’t depend on module parameters? var<private> decls? I think they should be duplicated)
+
+NOTE: aliases referring to the same declaration do not produce different concretizations.
+
+NOTE: each module concretization forms a distinct module. Importing multiple concretizations of the same module is allowed. 
+
+### Root Module: Shader Parameters
+
+Module parameters defined or published (TODO: publish spec) in the Root Module are called Shader Parameters. Shader Parameters can be set from the host code with a linker-specific API.
```
mighdoll, line 84: agreed, we'll need the publish spec in for this discussion of the root module to be complete

ModuleParameters.md
```diff
@@ -0,0 +1,100 @@
...
+
+NOTE: root module declarations are not reachable from other modules (due to cyclic imports) so shader parameters can only be replaced from the host code. (TODO: should we allow import cycles?)
```
mighdoll, line 86: I'd guess that circular imports are pretty common in TypeScript, at least for types. It's convenient for the programmer to not have to reorganize, and allowed by the compiler AFAIK. Perhaps module parameters will be the reason that we need to disallow that convenience?

ModuleParameters.md
```diff
@@ -0,0 +1,100 @@
...
+@param alias T = f32;
+@param struct Point { x: T, y: T }
+@param fn binop(lhs: T, rhs: T) {
+	return lhs + rhs;
+}
+```
+
+The importer module can decide to replace a module parameter with a compatible declaration (see Parameter Compatibility) in the import statement.
+
+Proposed syntax #1
```
mighdoll, line 22: Here `reduce_buffer` is a child of `binary_ops`? It's a little confusing..

ModuleParameters.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Syntax
+
+A module-scope declaration is preceded by the `@param` attribute (TODO or param keyword? choose a syntax) to turn it into a parameter. The initializer (or the function body) is the default value for the parameter, and is mandatory (TODO currently?).
+
+Examples
+```
+@param const MSAA_SAMPLES = 3;
+@param const ENABLE_XYZ = false;
+@param alias T = f32;
```
mighdoll, line 13: I think we should hold off on aliases as module parameters. It's a nice way to do generics! But I think we should leave this to future work until we decide on which approach to generics we like. If we end up with generic type parameters on interfaces, structs and functions, then having aliases as module parameters too may complicate the story for users and implementations.

ModuleParameters.md
```diff
@@ -0,0 +1,100 @@
+# Module Parameters
+
+Module parameters are module-scope declarations that can be overridden by other modules. constants, functions, structs and type aliases can be overridden. They allow a form of genericity and work hand-in-hand with conditional translation.
+
+## Syntax
+
+A module-scope declaration is preceded by the `@param` attribute (TODO or param keyword? choose a syntax) to turn it into a parameter. The initializer (or the function body) is the default value for the parameter, and is mandatory (TODO currently?).
```
mighdoll, line 7: module parameters is a nice concept, we are parameterizing the module here. 

re: word choice of `param`:

`override` is also nice, emphasizing that these will typically have a default value, but then there'd be both `override x = 1;` and `override const x =1;', which seems confusing.

other possibilities: `extern`, `configurable`, `open`.

ModuleParameters.md
```diff
@@ -0,0 +1,100 @@
...
+```
+import binary_ops with {binop: mul_f32}::{reduce_buffer as reduce_mul_f32};
+```
+
+Proposed syntax #3
+```
+@param(binop: mul_f32)
+import binary_ops::{reduce_buffer as reduce_mul_f32};
+```
+
```
mighdoll, line 37: perhaps:
```
import my_pkg::{foo, bar} with {ENABLE_XYZ: true};

fn foobar() -> f32 {
  return foo() + bar();
}
```

ModuleParameters.md
```diff
@@ -0,0 +1,100 @@
...
+A module-scope declaration is preceded by the `@param` attribute (TODO or param keyword? choose a syntax) to turn it into a parameter. The initializer (or the function body) is the default value for the parameter, and is mandatory (TODO currently?).
+
+Examples
+```
+@param const MSAA_SAMPLES = 3;
+@param const ENABLE_XYZ = false;
+@param alias T = f32;
+@param struct Point { x: T, y: T }
+@param fn binop(lhs: T, rhs: T) {
```
mighdoll, line 15: love the idea of param fn when we can inject from host code.. does it make sense to add param fn before then? perhaps! but I thought I should ask. perhaps a use case will make it clear.

ModuleParameters.md
```diff
@@ -0,0 +1,100 @@
...
+## Syntax
+
+A module-scope declaration is preceded by the `@param` attribute (TODO or param keyword? choose a syntax) to turn it into a parameter. The initializer (or the function body) is the default value for the parameter, and is mandatory (TODO currently?).
+
+Examples
+```
+@param const MSAA_SAMPLES = 3;
+@param const ENABLE_XYZ = false;
+@param alias T = f32;
+@param struct Point { x: T, y: T }
```
mighdoll, line 14: perhaps would defer this til we think through generics too. e.g. if we want to enable: 
```rs
fn foo<T>(t: T)
  where T: Point {
}
```
Then we'd have two potentially conflicting ways of doing a similar thing.

ModuleParameters.md
```diff
@@ -0,0 +1,100 @@
...
+  * If applicable, the original function’s return type is compatible with the replacement function’s return type (return contravariance)
+  * If the original function is void, the replacement function may return any type, or void.
+
+Examples: TODO
+
+## Behavior
+
+### Module concretizations
+
+Modules containing parameter declarations are called generic modules. Each usage of a generic module, with different parameters, forms a concretization of the module. Declarations inside different concretizations of the same module are duplicated. (TODO: what happens with declarations that don’t depend on module parameters? var<private> decls? I think they should be duplicated)
```
mighdoll, line 76: Here's a concrete example:

```rust
@param const aSize = 8;
@param const bSize = 8;

var<private> a: array<f32, aSize>;
var<private> b: array<u32, bSize>;

fn init_a() {
  for (var i = 0; i < aSize; i++) {
    a[i] = 0.0;
  }
}

fn init_b() {
  for (var i = 0; i < bSize; i++) {
    b[i] = 0.0;
  }
}

fn a_b() -> f32 {
  return a[0] + b[0];
}
```

ModuleParameters.md
```diff
@@ -0,0 +1,100 @@
...
+### Type Compatibility
+
+A type T2 is compatible with a type T1 if one of these conditions is met:
+
+* T1 equals T2
+* Transitively, T2 is compatible with T1 if U is compatible with T1 and T2 is compatible with U.
+* AbstractInt is compatible with AbstractFloat
+* f32 and f16 are compatible with AbstractFloat
+* i32 and u32 are compatible with AbstractInt
+* T1 and T2 are structs, T1’s fields exist in T2, and are each compatible with T1’s field type. T2 may have additional fields.
```
mighdoll, line 51: So structural typing for WESL structs? But WGSL is nominal, right? 

I think that implies that programmers could substitute T2 in a module parameter but not substitute T2 in a function call, which seems inconsistent.

ModuleParameters.md
k2d222: Yes, I agree. I tried to leave room for discussion and factoring in future options.

ModuleParameters.md
k2d222: Oh, in fact our import spec allows cyclic imports. I misremembered, and this is great :+1: 
So this NOTE is incorrect.

ModuleParameters.md
k2d222: yea, I sketched the example. We need a better one.

ModuleParameters.md
k2d222: Ok, so perhaps only consts and functions for now? I think we can drop structs and aliases while we figure out generics.

What about vars, which I didn't mention in the doc?

ModuleParameters.md
k2d222: Param is my favorite:
* Override is an overloaded term (btw: I think the `override` kw was very poorly chosen: it should be `overridable`!)
* "parameter" connects to generic parameters, which they are in some sense.
* I like how "module params" become "shader params" in the root module. It drops the concept of "conditionals/feature flags", which are simply shader params of type bool. The host interface mirrors the shader module interface.
* Not a fan of the other options, again some overloaded/confusing terms. "extern" would be my 2nd favorite, but quite far behind.

I prefer the attribute version, although I would not fight for it :)
Because I think this is just a good example of an attribute use: The code would be roughly the same without it; the attribute adds a detail to the declaration. "oh btw, this declaration can be replaced outside the module".
+ the advantage of not adding a kw and making it more obvious that it's a WESL extension.

ModuleParameters.md
mighdoll: consts sound good. var sounds fine for consistency, though I suppose any var could just be initialized with a @param const to get the equivalent functionality.

Function maybe - I think there's more to discuss on that: is there a use case we might discuss? the binops case is a bit complicated by generics.

ModuleParameters.md
mighdoll: I was warming to `open` as I wrote the above. Though I like `param` too. We can ask the webgpu folks for opinions on whether to `@` if you'd like. Or survey on social media.

> drops the concept of "conditionals/feature flags", which are simply shader params of type bool
love this!

ModuleParameters.md
k2d222: It's a good one, having the `with` at the end reduces clutter.

@ncthbrt recommended `=` sign instead of `:`, as the latter is for types and the former for assigns. Which I agree with.

I think there is no good-looking solution when the generic module a middle path component. One option is to make the replacement apply to any module in the path that has the param.

For completeness here are all the variations:
(my favorites are 3b and 2b)

```js
// (1) using angle brackets <>
// ---------------------------

// (1.a) param1 applied to generic_mod only
//       I find the mix of :: and <> hard to read
import some::generic_mod<param1=value>::child_mod::{foo};

// ... or with turbofish operator ::<>
import some::generic_mod::<param1=value>::child_mod::{foo}; 

import some::generic_mod<
    param1=value,
    param2=value,
>::child_mod::{foo};


// (1.b) param1 applied to all modules in path
//       reminds of var<storage>, separates :: and <>
import<param1=value> some::generic_mod::child_mod::{foo};

import<
    param1=value,
    param2=value,
> some::generic_mod::child_mod::{foo};


// (2) using with { .. }
// ---------------------------

// (2.a) param1 applied to generic_mod only
//       is quite cluttered and { } confusion
import some::generic_mod with { param1=value }::child_mod::{foo};

import some::generic_mod with {
    param1=value,
    param2=value,
}::child_mod::{foo};

// (2.b) param1 applied to all modules in path
//       more readable, especially on new line
import some::generic_mod::child_mod::{foo} with { param1 = value };

import some::generic_mod::child_mod::{foo} with {
    param1 = value,
    param2 = value
};

import some::generic_mod::child_mod::{
    foo,
    bar,
} with {
    param1 = value,
    param2 = value
};



// (3) using @param( .. )
// ---------------------------

// (3.a) param1 applied to generic_mod only
//       very ugly & unreadable
import some::@param(param1=value) generic_mod::child_mod::{foo};

import some::@param(
    param1=value,
    param2=value,
) generic_mod::child_mod::{foo};

// (3.b) param1 applied to all modules in path
//       nice separation between import path and params
@param(param1=value)
import some::generic_mod::child_mod::{foo};

@param(
    param1=value,
    param2=value,
)
import some::generic_mod::child_mod::{foo};
```

ModuleParameters.md
k2d222: Yes it's a bit suspicious and perhaps we should not allow param structs.
There is definitely a need for "extending structs", and this would be a way to do it.

As the param is a kind of generic parameter, the default struct acts more like a structural constraint. The function can be called only with one concrete type once the module is imported (the "generic is instanciated").

mighdoll: I like being able to extend structs too, though as I say that I think we should have some key examples handy to consider. 

I used a variant of the struct extension idea in the pre-wesl linker to allow libraries / modules to contribute fields for the uniforms buffer. The idea was that modules / libraries would specify what fields they care about `MyLib_Uniforms` and the app shader would construct a single composite uniform struct from the libraries' and modules' uniforms. Perhaps that idea might still make sense, but we may have other ideas now on how to solve that underlying problem of passing uniform data from host code to library code.

ModuleParameters.md
mighdoll: Do we need to set module parameters for middle path components? I think import only brings the final leaves of the import statement tree into scope.. from our wesl spec:

> Once the import has been resolved, the last segment, or its alias, is brought into scope.

..So I'm not yet seeing quite yet why we need to set module path parameters for the middle path segments. can you straighten me out?

> @ncthbrt recommended = sign instead of :, as the latter is for types and the former for assigns. 

Good fix. (`':'` looks natural in ts/js, but I agree that a consistent rule for wesl/wgsl is better.)

ModuleParameters.md
mighdoll: also on function module params, perhaps we should consider interaction with function pointers/references. e.g. if we want to have const function references:

```rs
@param fn foo(s: MyStruct) -> u32 { }
const f: (s: MyStruct) -> u32 = foo;

fn bar() { zing(f); } 
```

function references seem useful when we're selecting which function to use based on conditions, and we want to pass that decision around. 

the question is what happens to to the type of `f` if `foo` is replaced with a compatible but different struct type, e.g. s: MyStruct + MyEnhancement.

In the const fn pointer case, we'd convert the function pointers to direct calls in transpilation. 

Dynamic dispatch would raise similar issues if we ever support dynamic. (the slang folks argue that dynamic dispatch is valuable for modern rendering pipelines.)

I'm not convinced we want function references, but if you agree it's reasonably likely, then we should consider how they would mix with function module params.

ModuleParameters.md
ncthbrt: Would extending structs not benefit from having an extends keyword that allows users to control where the fields from the base struct are added?

For example:

```rescript

struct A {
    a: f32;
    b: f32;
}

struct B {
    c: f32;
    extends A;
}
```

stefnotch: What would be use cases for structs as module parameters?

Depending on the answer, I might want to "use a struct for multiple modules". In which case one would get multiple inheritance, and all the associated pains.

That is to say, maybe we should really be separating between "behaviour (interfaces)" and "memory layout (structs)"?

ModuleParameters.md
ncthbrt: I think we should start to consider structural interface types for this stuff. Like foo should take in specifically a concrete MyStruct but a generic function could specify an interface structural type

ModuleParameters.md
```diff
@@ -0,0 +1,100 @@
+# Module Parameters
+
+Module parameters are module-scope declarations that can be overridden by other modules. constants, functions, structs and type aliases can be overridden. They allow a form of genericity and work hand-in-hand with conditional translation.
```
mighdoll, line 3: I suppose you can set module module parameters based on a const in your own module? Even a param const in your module that was set from host code.

mighdoll: (but not a from a var of course)

ModuleParameters.md
mighdoll, line 1: I'm wondering if the fork-module-on-write approach will cause confusion and if so what we do about it.
Say module A imports from module B and sets a param const. Then, say the root shader app imports directly from module A and module B. The root shader would see two copies of module B. This seems a risk for causing confusion.. 

perhaps we ought only allow one version of a given module rather than forking a new concrete instantiation for each variation? (we'd error if a param const was set to two different values.)

In this example, there's two buffers, which might be unexpected. start() is 0.
but if we change the defaultSize to 4, there's one buffer and start() is 1.

```rs
/// --------   File: app.wesl -----------
import package::A::sum;
import package::B::first;

fn start() { init(); return first(); }

/// --------   File: A.wesl -----------
import package::B::buf with {bufSize=defaultSize};

const defaultSize = 8;

fn init() {
  for (var i = 0; i < maxElems; i++) { buf[i] = i; }
}

/// --------   File: B.wesl -----------
@param const bufSize = 4;
var buf<workgroup> array<f32, bufSize>;

fn first() -> f32 {
  return buf[0];
}

```

stefnotch: (Good question. I suspect that module parameters might fundamentally be the wrong tool for most of our current issues.)

mighdoll: > (Good question. I suspect that module parameters might fundamentally be the wrong tool for most of our current issues.)

What if we narrow to just `@param` consts, and don't fork on write. Is that narrower view a clean design? I like that it would get us some nice improvements: 
- you could set condition variables from shader code,
- more IDE friendly and more typechecked runtime constants when we pass them from host code.
- unification of runtime conditions and runtime consts

stefnotch: We would still have fork-on-write with constants that can be re-assigned, right?
Otherwise we'd get constants that can be re-assigned exactly once, which seems like a very suspicious language feature.

I'm okay with implementing this if it solves some immediate issues, but I don't expect it to be the proper long term solution.

mighdoll: Right, I'm floating the idea that we'd allow assigning to param consts only one new value. It's a different and more narrow conception. The narrow view leaves the abstract interface problems of [monoid-sketch](https://github.com/wgsl-tooling-wg/monoid-sketch) undecided and hopefully unaffected by this.

I like that it would still solve several issues, #131 and #132 among them. Also gives us a spec'd way to handle our current linker generated consts that's more typechecked and ide supportable. (and might even help to allow for binding indices in libraries..)

But perhaps your glimpsing a different long term solution?

k2d222: IMO modules imported with different params are conceptually distinct modules (they are "module generators") and I don't see the fork being a big issue in practice. I mean, every decent programmer should know that mutable global state can go boom.

To mitigate that we could make param consts always mandatory (no default value)

With that said, if you guys think it's a serious issue, then allowing only 1 instance is definitely not the right solution and I'd rather we drop the proposal altogether. There are plenty of ideas out there, let's keep digging :)

stefnotch: I'm envisioning a shiny future where conditions are just "const generics of type bool". We already have that in WGSL's type system, so it'd be a very sensible addition.

```rust
fn main<MOBILE: bool>(...) -> ... {
  const if(MOBILE) {
    // run mobile code
  } else {
    // other code
  }
}
```

And then we have new problems to solve
- Manually passing `MOBILE` as a generic argument to many functions gets old
- Adding a group of constants is frustrating https://github.com/wgsl-tooling-wg/wesl-spec/issues/132

And to solve that, we extend the generics with a "context". Now one can reasonably pass the constants to where they need to go, without adding extra code at the call sites. https://github.com/wgsl-tooling-wg/monoid-sketch/blob/main/context.wesl

mighdoll: I'm interested in continuing to explore the param const possibility. I think a little more digging will reveal whether it's the right path.

Multiple forks from importing with different `@param const` values could indeed be ok.. The example I raised was a bit contrived. And we could later add some syntax / ide support. e.g. warn when there are multiple versions unless the user imports with an exclamation point. `import package:buf with! { bSize = 8};`.

But explain more about why the single variant limitation isn't worthy? Admittedly, if we stop there, it's not a module constructor, just a module customizer. And that's a different conception from where we started. But a module customizer seems a pretty natural solution to cleaning up a bunch of stuff with runtime consts and conditions. And if we think a module customizer is a good design, we can later relax the restriction and allow multiple variants.

mighdoll: > ```rust
> fn main<MOBILE: bool>(...) -> ... {
>   const if(MOBILE) {
>     // run mobile code
>   } else {
>     // other code
>   }
> }
> ```

More powerful function handling seems like a good direction. And a context mechanism is pretty tempting too (I noticed that LEAN has implicit params, ocaml is trying to retrofit them, kotlin, scala, have 'em, etc. The idea seems popular among high class language designers). 

However, conditions and constants can be at the root level of the file with current wgsl too, not just in functions. We could start requiring that conditions and param constants always be inside some new construct we add (a trait or module or whatever). Then we could condition and parameterize. But that requirement would mean moving a bit further away from wgsl, which is undesirable if we can avoid it..

k2d222: @mighdoll Re: single variant limitation isn't worthy

On a second thought it might be. I'm just afraid that there might be a "action at a distance" footgun, because changing a const value somewhere may cause a module import incompatibility elsewhere.

I like that condcomp params are no longer undefined globals, plus the ability to define them from shader code. So it might be worth it. 

WDYT of the "param const always uninitialized and mandatory" variant? To make explicit what "generic module" is constructed at import time.

mighdoll: > I'm just afraid that there might be a "action at a distance" footgun, because changing a const value somewhere may cause a module import incompatibility elsewhere.
>
> WDYT of the "param const always uninitialized and mandatory" variant? To make explicit what "generic module" is constructed at import time.

It's a good idea. I'm generally pretty in favor of starting limited and then relaxing constraints.   

The other side of the balance is the convenience of defaults, especially when the const value is only rarely of concern externally. 

In this case, I'd be pretty tempted to allow both uninitialized and initialized consts at first. If it becomes a footgun in practice we'll hear about it. But if we don't allow for initialized param consts, the use cases that make sense with defaults will remain hidden.

mighdoll:
Is it reasonable to use 'override' instead of `param const`? 
- An unpublished `override` in WESL becomes a `const` in WGSL (i.e. an `override` not imported into the root module with a `publish`). So they're pretty similar.
- But an `override` in wgsl comes with additional power. Unlike a `const`, an `override` can be modified in the shader module from host code _after_ linking and shader module creation. 
- And `override`s have additional limitations (e.g. can't be used for binding indices).
- And `param const` has the additional power that they can be used in conditionals, which isn't the case for late modifiable `override`.

We'd talked about trying to make `param const` and `override` closer syntactically. 
- `override` and `override const` are close syntactically, but imply that `override const` has the power to host modified late like `override` which isn't true.
- maybe there's some other syntax that makes `override` a special version of `param const`? `@override param const`?

ModuleParameters.md
ncthbrt: I think injection should rather be done through a formal generics implementation tbh. This feature is basically this but less integrated

ModuleParameters.md
k2d222: How about this: Every parameterized module import creates a new, isolated module. So importing the same module twice, even with the same parameters, copies all local variables.

This form works similarly to the Functor design: Calling the functor several time creates multiple distinct "instances" of the module.

I see some issues with this approach, but I think it might work too.

k2d222:
Re @mighdoll override:

I hope we don't settle on `override const` or `@override param const`, which seem too convoluted convoluted (reminds me of the [provocative `var const` in DreamBerd](https://github.com/TodePond/GulfOfMexico))

I like the idea of reusing existing concepts. But in this case, I feel that the conceptual overlap between pipeline-overrides and module-overrides is too distinct. A user might naturally think "oh, this is an override so I can change it at runtime". But as you mentionned, the "module-overrides" could be part of conditionals and binding indices.

Also I believe that `override` was a poorly chosen name for WGSL: it's short for `overridable`, which would have been a fitting name; and `override` has different meanings in other langs.

ModuleParameters.md
ncthbrt: I think that may be quite confusing behaviour and would immediately rule out adding inline module usage (meaning aliasing would be much more necessary). I would suggest we start with specialising once and then add a keyword (like `new`) to create a new instance of a module

mighdoll: For the simple examples, seems like specializing once for M2 is ok. Or are there are some cases that make clear that the just once limit will be awkward? 

Generally, hoping we can deliver some good improvements with M2 that people can try, and we can expand later. We can't settle typeclass style vs. module style vs. whatever style direction in M2 of course, but param const functionality seems like it'll fit right in, maybe with some syntax tweaks in M3.

k2d222:
Resolution from last meeting: Make a new PR with a narrower scope:
- only param consts, no param functions, structs etc.
- can be set only once
- only set from the host code, for now.
  - what about packages? can we set param consts of dependencies in `wesl.toml`?
- possible extension: perform arithmetic in `@if` using param consts (e.g. `@if(NUM_BINDINGS > 3)`)
  - it's relatively simple since param consts are known by the host.

PR #147, opened 2025-08-26
Inline Functions Proposal

k2d222:
Again, rough draft and see TODOs for comments.

### DIFF
--- a/InlineFunctions.md
+++ b/InlineFunctions.md
@@ -1,0 +1,108 @@
+# Inline Functions
+
+Inline functions are inlined at the calling site, meaning the function body is “copied” in place of the call expression. Inline functions have both advantages and limitations compared to typical functions.
+
+## Syntax
+
+Function declarations preceded with the `@inline` attribute are marked as inline functions.
+
+## Behavior
+
+Inline functions behave somewhat like macros, in the sense that their contents are inlined at the call site, and the declaration is erased at compilation time. 
+
+However, inline functions are hygienic, meaning that they can only access declarations in scope. In-scope declarations are module-scope declarations and function parameters.
+
+(TODO) should we allow discard statements? diagnostics? what should be forbidden in inline functions?
+
+### Parameter and Return Types
+
+Inline functions lift some limitations on allowed parameters and return types. In WGSL, a function return type must be constructible. The return type of an inline function can be a constructible type, a texture, sampler, or pointer type. An inline function may also return a parameter or a module-scope declaration, including bindings. Returning parameters or module-scope declarations does not perform a copy, instead they must are “returned by reference” and can only be bound to a variable.
+
+In WGSL, function-scope var-declarations must have a constructible type. Exceptionally, non-constructible types returned from inline functions can be bound to a var-declaration. 
+The scope, address space and access mode of the declaration is inferred from the inline function return value. (TODO: should we allow explicit? in particular the access mode, if applicable.)
+Non-constructible types cannot be bound to value declarations.
+
+NOTE: In inline functions returning non-constructible types, the code path leading to the return statement cannot contain branches, so the returned value is determined at shader-creation time. But the function can contain runtime side-effects alongside.
+
+Inline functions can call other inline functions, as well as non-inline functions.
+
+### Limitations
+
+Entry-points functions cannot be inline.
+
+Inline functions can only be invoked from inside a function body.
+(TODO: an inline function containing a single return statement could be inlined in const-expressions. Though, this is limited too and could be a job better suited to user-defined `@const` functions)
+
+Inline functions with a return value *must* contain a single return statement at the end of the function body (the last statement), and *must not* contain any early return. This rule is only checked after full resolution of conditional translation attributes.
+(TODO can we lift this limitation? we could copy-paste the code following the function invocation in all return branches. This could lead to unexpectedly large code, and unexpected non-uniform control flow. But is much more powerful otherwise.)
+
+NOTE: This ensures that the macro outer control flow has no branch and can be inlined without affecting the following code.
+
+NOTE: Compile-time branching can be performed inside inline functions with the help of conditional translation. See the example below.
+
+## Implementation
+
+### Validation
+
+The implementation *should* check that the returned value from the inline function matches the function return type. Otherwise, the code generation may produce valid WGSL code from invalid WESL code.
+
+The implementation *must* ensure that non-constructible returned values can only be bound to var-declarations. The var-declaration must not contain an explicit address space or access mode. It is inferred. (TODO)
+
+Other cases of invalid use of inline functions would generate invalid WGSL code, therefore validation is optional.
+
+### Code Generation
+
+Implementations are free to generate code differently. This is an example of a conforming code transformation.
+
+1. Rename all identifiers declared in the inline function. See Name Mangling.
+2. Insert the inline function statements before the statement containing the call expression, ignoring the return statement.
+3. 
+  a. If the inline function returns an identifier referring to a module-scope declaration, or a parameter to the inline function, and the call expression is part of a var-declaration: The var-declaration is removed, and all references to the declaration are renamed to the returned identifier.
+  b. Otherwise, replace the call expression with the returned expression.
+
+NOTE: If the function is called from a call statement: a simpler code generation is to replace the call statement with a compound statement containing the function body.
+
+### Name Mangling
+
+When renaming identifiers, the implementation is free to pick any name that does not conflict with identifiers in scope at the call site.
+
+We recommend the following renaming scheme for identifiers declared in inline functions:
+
+(TODO)
+
+## Examples
+
+(TODO)
+
+The following snippet:
+
+```
+@group(0) @binding(0) sampler_a: sampler;
+@group(0) @binding(1) sampler_b: sampler;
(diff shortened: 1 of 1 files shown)


InlineFunctions.md
```diff
@@ -0,0 +1,108 @@
...
+Inline functions behave somewhat like macros, in the sense that their contents are inlined at the call site, and the declaration is erased at compilation time. 
+
+However, inline functions are hygienic, meaning that they can only access declarations in scope. In-scope declarations are module-scope declarations and function parameters.
+
+(TODO) should we allow discard statements? diagnostics? what should be forbidden in inline functions?
+
+### Parameter and Return Types
+
+Inline functions lift some limitations on allowed parameters and return types. In WGSL, a function return type must be constructible. The return type of an inline function can be a constructible type, a texture, sampler, or pointer type. An inline function may also return a parameter or a module-scope declaration, including bindings. Returning parameters or module-scope declarations does not perform a copy, instead they must are “returned by reference” and can only be bound to a variable.
+
```
mighdoll, line 20: can the inline function call be a parameter to a function call?

InlineFunctions.md
```diff
@@ -0,0 +1,108 @@
...
+Function declarations preceded with the `@inline` attribute are marked as inline functions.
+
+## Behavior
+
+Inline functions behave somewhat like macros, in the sense that their contents are inlined at the call site, and the declaration is erased at compilation time. 
+
+However, inline functions are hygienic, meaning that they can only access declarations in scope. In-scope declarations are module-scope declarations and function parameters.
+
+(TODO) should we allow discard statements? diagnostics? what should be forbidden in inline functions?
```
mighdoll, line 15: disallowing discard and const_assert and diagnostics sounds like a good place to start..

stefnotch: If we only allow for them to be called from other functions, could we scope the diagnosics by inserting a compound statement? `{ }`

What happens when one calls discard in a normal function?

Also, wesl implementations will have to check for `break` and `break if`s. Otherwise I could sneak illegal code past a WESL compiler.
```rust
@inline
fn b() { break; } // very illegal

fn main() {
  while 1 == 1 {
    b();
  }
}
```

InlineFunctions.md
```diff
@@ -0,0 +1,108 @@
...
+
+Inline functions lift some limitations on allowed parameters and return types. In WGSL, a function return type must be constructible. The return type of an inline function can be a constructible type, a texture, sampler, or pointer type. An inline function may also return a parameter or a module-scope declaration, including bindings. Returning parameters or module-scope declarations does not perform a copy, instead they must are “returned by reference” and can only be bound to a variable.
+
+In WGSL, function-scope var-declarations must have a constructible type. Exceptionally, non-constructible types returned from inline functions can be bound to a var-declaration. 
+The scope, address space and access mode of the declaration is inferred from the inline function return value. (TODO: should we allow explicit? in particular the access mode, if applicable.)
+Non-constructible types cannot be bound to value declarations.
+
+NOTE: In inline functions returning non-constructible types, the code path leading to the return statement cannot contain branches, so the returned value is determined at shader-creation time. But the function can contain runtime side-effects alongside.
```
stefnotch, line 25: Do we really need this restriction? 

I thought I could write stuff like
```
var a: ptr<...> = &foo;
if something {
  a = &bar;
}
```

It was just problematic for textures because of nom-uniform control flow leading to missing partial derivatives when sampling.

ncthbrt:
Another idea I had related to this is yield blocks...

What if you could `yield` code blocks inside a const fn (sort of similarish to an enumerable in c#)? I guess that's starting to become macro territory then though...

InlineFunctions.md
k2d222: yes. There is one limitation though, that I didn't write down: order of evaluation.

```rust
let x = foo(sideEffect(), inlineFunc());
```

If `inlineFunc` has statements, they will be inlined before the `x` declaration statement, so before the `sideEffect`.
```rs
{ /* inline inlineFunc's side-effects */ }
let x = foo(sideEffect(), /* inlineFunc's return */);
```

Ofc a more complex codegen can circumvent this:

```rs
let arg1 = sideEffect();
{ /* inline inlineFunc's side-effects */ }
let arg2 = /* inlineFunc's return */;
let x = foo(arg1, arg2);
```

InlineFunctions.md
k2d222: Your code would be valid and I misphrased the sentence. It should have been
> NOTE: Since there can be only one return statement AND non-constructible types are not assignable, the returned value cannot be depend on a runtime condition, so it can be determined at shader-creation time.

InlineFunctions.md
```diff
@@ -0,0 +1,108 @@
...
+```
+@group(0) @binding(0) sampler_a: sampler;
+@group(0) @binding(1) sampler_b: sampler;
+// …
+
+fn main() {
+    // …
+    textureSample(myTexture, sampler_a, myCoords);
+}
+```
```
stefnotch, line 108: What would the transformation look like in more complex cases?

I'm guessing this
```rust
@inline
fn selectSampler() -> sampler { return sampler_a; }

fn main() {
    // …
    var mySampler = selectSampler();
    textureSample(myTexture, mySampler, myCoords);
    textureSample(myOtherTexture, mySampler, myCoords);
}
```
needs to be transformed into
```rust
fn main() {
    textureSample(myTexture, sampler_a, myCoords);
    textureSample(myOtherTexture, sampler_a, myCoords);
}
```
because we cannot store a sampler in a local variable. See also https://github.com/gpuweb/gpuweb/issues/2482

With that in mind, *what data type does `mySampler` have?*

mighdoll: would it not have the type of sampler_a? e.g. `sampler`

I'd guess we should disallow a var. But let's say for a let.

Are you saying then we should also allow `sampler` after the colon in a let statement inside a fn? 

```rs
    let mySampler: sampler = selectSampler();
```

stefnotch: I'm asking, because that `let mySampler` variable needs to be removed during lowering. WGSL currently does not allow `let mySampler = someSampler`.

stefnotch:
I suspect that this `@inline` doesn't actually behave how I'd expect an `@inline` to behave.

Case in point, the following code
```rust
var<storage, read_write> counter: atomic<u32>;

@inline
fn next_random() -> u32 {
  return atomicAdd(&counter, 1);
}

fn do_something(value: u32) { ... }

fn main() {
  let a = next_random();

  do_something(a);
  do_something(a);
  do_something(a);
}
```

would turn into
```rust
var<storage, read_write> counter: atomic<u32>;

fn do_something(value: u32) { ... }

fn main() {
  do_something(atomicAdd(&counter, 1));
  do_something(atomicAdd(&counter, 1));
  do_something(atomicAdd(&counter, 1));
}
```

But why?
Well, because that's how our inline desugaring *has to* work for textures, samplers, etc. Take the following code.
```rust
var<storage, read_write> counter: atomic<u32>;

@inline
fn get_texture_with_side_effect() -> texture_2d<f32> {
  atomicAdd(&counter, 1);
  return my_texture;
}

fn do_something(value: texture_2d<f32>) { ... }

fn main() {
  let a = get_texture_with_side_effect();

  do_something(a); // increments
  do_something(a);  // the counter
  do_something(a);  // 3 times
}
```

stefnotch:
Good to hear that you already thought about this 👍

k2d222:
I didn't properly think of this, It just happens to be a happy coincidence. Thanks for finding the cracks!

k2d222:
I new idea that would resemble bindless. Instead of inlining, we rewrite the variable as a `u32` index that refers to one of the possible instances (all instances of the type get assigned a unique ID). Later usage of the variable gets desugared as a switch-case. 

In the example below, I use a `&type` syntax to indicate that we are dealing with a "reference to" the type, and `*type` where I want to use the instance. The nice thing is that the function is not inlined, the inlining actually happens at the usage site, where I use `*type`.

Example:

```rs
fn select_texture(tex1: &texture_2d<f32>, tex2: &texture_2d<f32>, some_arg: f32) -> &texture_2d<f32> {
    if (some_arg > some_threshold) { return tex1 } else { return tex2 };
}

// function above gets immediately rewritten as:
fn select_texture(tex1: u32, tex2: u32, some_arg: f32) -> u32 {
    if (some_arg > some_threshold) { return tex1 } else { return tex2 };
}

fn main() {
    let tex: &texture_2d<f32> = select_texture(&myTex1, &myTex2, myCond);
    let texel = textureLoad(*tex, coords, 0);
}

// function above gets rewritten as:
fn main() {
    let tex: u32 = select_texture(0, 1, myCond);
    
    var _temp;
    switch (tex) {
        case 0: { _temp = textureLoad(myTex1, coords, 0); }
        case 1: { _temp = textureLoad(myTex2, coords, 0); }
        default: { _temp = /* some default, unreachable value */; }
    }
    let texel = _temp;
}
```

I think this solution is strictly better for the example above of selecting textures / samplers:

* the codegen is rather simple and predictable. The programmer knows that at the dereference site there will be a switch-case, and the function call remains a function call.
* in contrast to the `@inline` approach, it allows runtime branching!
* ...and if the linker is a bit smart, it can make it zero-cost if branching is compile-time.
* it's an implementation of the bindless paradigm, which can be otherwise awkward to implement with explicit indices.

PR #150, opened 2025-09-08
fix for 'common' is too common - relax module path segment syntax

### LINKED ISSUE #127: 'common' is too common - relax module path segment syntax
The current WESL syntax forbids keywords and reserved words in import module paths. But Lygia uses 'common', 'mod', and 'const' for shader files and directories. All of those can't be imported in WESL because they're keywords or reserved words in WGSL. 

It seems too easy to pick a file or directory name which happens to conflict. WGSL has an expansive list of reserved words, and users aren't accustomed to restricting their file and directory names. Let's consider allowing keywords and reserved words in module paths.

mighdoll:
Fixes #127

### DIFF
--- a/Imports.md
+++ b/Imports.md
@@ -78,7 +78,12 @@ import_collection:
 ```
 
 Where `translation_unit` and `ident` are defined in the WGSL grammar.
-`ident`s are subject to the usual restrictions, meaning they cannot be keywords or reserved words.
+`ident`s must not be current WGSL keywords. `ident`s also must not be 
+current WESL keywords: `as`, `import`, `package`, `super`, or `self`. 
+Reserved words that are
+not current keywords are allowed, 
+but not recommended.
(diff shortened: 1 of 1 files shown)


Imports.md
```diff
@@ -78,7 +78,12 @@ import_collection:
 ```
 
 Where `translation_unit` and `ident` are defined in the WGSL grammar.
-`ident`s are subject to the usual restrictions, meaning they cannot be keywords or reserved words.
+`ident`s must not be current WGSL keywords. `ident`s also must not be 
+current WESL keywords: `as`, `import`, `package`, `super`, or `self`.
```
stefnotch, line 82: Could you disallow all current keywords instead of listing them?

stefnotch: Ah, that's what you did. Those are just WESL keywords.

k2d222:
Approved.

PR #164, opened 2025-11-13
Module parameters proposal, take 2

### LINKED ISSUE #117: @else into spec
`@else` has been user requested and seems pretty natural every time we try write examples too. Shall we add it to the spec? It's currently a [possible future extension](https://wesl-lang.dev/spec/ConditionalTranslation#possible-future-extensions)

k2d222:
supersedes #146. 
closed #117.

Greatly simplified as commented in https://github.com/wgsl-tooling-wg/wesl-spec/pull/146#issuecomment-3466771471.

- `@param const` create overridable module parameters
- can be set only once, by the linker only (for now? what about packages? imports?)
- can be used in `@if` (replaces *conditional features*)

I also added `@elif` and `@else` in the spec (#117)

[Rendered](https://github.com/wgsl-tooling-wg/wesl-spec/blob/k2d222-patch-1/ConditionalTranslation.md)


mighdoll (commented):
Perhaps we should also add numeric `@param` const values in this revision? even w/o using them in conditions, that gets us out of the nonstandard `constants::` we're doing now..

mighdoll:
I note that `param` isn't currently a reserved word in WGSL, which would make `param` syntax harder to adopt upstream..

k2d222:
@stefnotch 
I think I `@param const` because there is already a lot of declaration keywords – var, var<>, let, const, override and it's confusing. Also I would prefer if all consts could be used in `@if`, but the implementation cost of const eval doesn't justify it yet.

Thinking about it, I would like to amend the proposal: "all consts initialized with a simple literal can be used in `@if`, while `@param` just makes the const host-visible. Or even, we ditch `@param` and use `@public` instead.

@mighdoll 
love the table and diagram! makes things clear right away.

Ok to add numeric params in this revision.

stefnotch:
Very relevant here: Section "4.2.6 Global Generic Type Parameters" from the Slang paper https://kilthub.cmu.edu/articles/thesis/Slang_A_Shader_Compilation_System_for_Extensible_Real-Time_Shading/16826602?file=31117213

I do encourage reading that section. A rough summary of it is:

Slang has generic entry points. They let you specialize a shader.
```hlsl
float3 entryPoint<M : IMaterial, L : ILightingEnv>(
  Camera camera,
  M material,
  L lights,
  SurfaceGeometry geometry)
{
  float3 viewDir = normalize(camera.P - geometry.P);
  M.Pattern bxdf = material.evalPattern(geometry);
  return lights.illuminate(bxdf, viewDir);
}
```
but there's a lot of existing HLSL code.
> It is common practice in current HLSL and GLSL codebases to declare some of the parameters of  a shader entry point at global scope, rather than as explicit parameters of an entry-point function

Some of it was hard to convert. So they added a mechanism for that.
That mechanism is essentially our param const, except it's for types instead of values.
```hlsl
type_param M : IMaterial;
type_param L : ILightingEnv;
ParameterBlock<M> gMaterial;
ParameterBlock<L> gLights;
ParameterBlock<Camera> gCamera;
float4 main(SurfaceGeometry geom)
{
  float3 viewDir = normalize(gCamera.P - geometry.P);
  M.BxDF bxdf = gMaterial.evalPattern(geometry);
  return gLights.illuminate(surface, viewDir);
}
```
> Conceptually they behave as if the whole shader library  is nested in a generic declaration, with the given parameters

And then they link to section 7.2.3 where they explain why the `type_param` design is worse. Based on their experience in shipping production shader codebases.
> In practice, global generic type parameters have created many problems and should be avoided if possible.

The problem there is that one runs into "Confusing circular dependency scenarios" when one starts doing simple kinds of reflection.

To create a shader with `type_param M`, you have to supply a type. Often times, you'll want to use some type that is defined in your shaders. *But*, that type is not allowed to depend on `M` in any way. `M` doesn't exist yet after all.

The param const equivalent is
```wgsl
param const X: u32;
const Y = X + 1;
```
and then trying to read the value of `Y` via reflection. You first have to pass in the param const values to be able to create the shader.

In HLSL/Slang, this problem is much worse, since they have methods directly in their struct definitions. So here, I couldn't use reflection to read `D` until after I've passed in something for `T`.
```hlsl
type_param T;
struct D
{
  float doSomething()
  {
    T t; ...
  }
};
```

stefnotch:
I think we can defer it until we have some generics prototypes. Unless there are some pressing use-cases that this would unblock?

k2d222:
It would be nice to have it before bevy adoption of WESL. Because 
* this proposal is a breaking change for conditional compilation
* bevy uses numeric conditions

Otherwise I'm happy to defer.

k2d222:
> Root module: Param consts are host visible, so the root module controls them. When we add "publishing/public", then we'll extend the permitted usages from "only param consts in the root module are visible" to "root module and public items are visible"

Does that mean that param consts in non-root modules are effectively consts for the time being? Should we just forbid them?

> Setting from shader code: Do we have users that immediately need this, or can we defer that can of worms?

can defer I believe.

PR #168, opened 2025-11-24
Updates to wesl toml semantics

stefnotch:
We talked about this on Discord, this just notes down what the decisions were

### DIFF
--- a/WeslToml.md
+++ b/WeslToml.md
@@ -72,6 +72,11 @@ IDEs watch this directory for changes.
 
 - Optional
   - Defaults to the `shaders` directory adjacent to the `wesl.toml`. (`"./shaders/"`)
+
+When the `wesl.toml` file does not exist, then tools should choose one of the following behaviours
+- Erroring out, for tools that rely on `wesl.toml` features. For example, an invocation `wesl fmt` would do this.
+- Accepting another way of configuring where the `root` is.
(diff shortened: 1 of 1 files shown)


WeslToml.md
```diff
@@ -72,6 +72,11 @@ IDEs watch this directory for changes.
 
 - Optional
   - Defaults to the `shaders` directory adjacent to the `wesl.toml`. (`"./shaders/"`)
+
+When the `wesl.toml` file does not exist, then tools should choose one of the following behaviors
+- Erroring out, for tools that rely on `wesl.toml` features. For example, an invocation `wesl fmt` would do this.
+- Accepting another way of configuring where the `root` is.
+- Assuming that a default `wesl.toml` file is right next to the `package.json`/`Cargo.toml`/`deno.json`. This is what the language server will do.
```
mighdoll, line 79: Shouldn't all compliant tools (e.g. wesl fmt) do what the lang server does here and presume a default wesl.toml if it's missing?

stefnotch: I should have added more examples.

> Erroring out

That's the easy option for tools that don't want to maintain a list of accepted project files. (`package.json`/`Cargo.toml`/`deno.json`)
And for tools where the scanning behavior might be confusing. 

> Accepting another way of configuring where the `root` is.

That was written with the intention of allowing 
`wesl fmt typegpu/src/shaders` or `wesl fmt myshader.wesl`.

Though the phrasing is bad. Hmm.

PR #171, opened 2025-11-29
Allow `@if` on assignment and increment statements

k2d222:
See https://github.com/wgsl-tooling-wg/wesl-rs/issues/162 and https://discord.com/channels/1275293995152703488/1275293995152703491/1443216184001691688 discussions. Now all statements support attributes; assignment was the "odd one out".

This change breaks the LR grammar, but the benefit is worth it.

### DIFF
--- a/ConditionalTranslation.md
+++ b/ConditionalTranslation.md
@@ -11,7 +11,7 @@ This attribute indicates that the syntax node it decorates can be removed by the
 
 ### Usage Example
 
-```wesl
+```wgsl
 // global variables and bindings...
 @if(textured)
 @group(0) @binding(0) var my_texture: texture_2d<f32>;
@@ -35,7 +35,7 @@ fn main() -> vec4 {
 
 Quirky examples
 
-```wesl
+```wgsl
 // attribute order does not matter.
 @compute @if(feature) fn main() { }
 // ... is equivalent to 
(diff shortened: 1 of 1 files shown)


ConditionalTranslation.md
```diff
@@ -114,13 +113,18 @@ Refer to the [updated grammar appendix](#appendix-updated-grammar) for the list
    * function call statements
    * const assertion statements
 
+> [!WARNING]
+> We added attributes on assignment and increment/decrement statements. This change makes the WESL grammar no longer [LR(1)](https://en.wikipedia.org/wiki/LR_parser). See [`wesl-rs#162`](https://github.com/wgsl-tooling-wg/wesl-rs/issues/162) for details.
```
stefnotch, line 117: I know too little about the different kinds of parsers. Is wesl still LALR(1), or does it also break that?

k2d222: ~~It's still LR, (or LALR),~~ yes, it breaks LALR or LR(1) parsers. And the workaround is that `@if(foo) (x)++` is not supported, and it's impossible to do so with a LR(1) parser. But `(x)++` alone is OK and so is `@if(foo) x++`. So not a big deal, we are quite far in edge-case territory.

k2d222: clarification: LALR is a less expressive version of LR(1), but very similar in practice. So yes, this change makes WESL no longer LR(1), or LALR. Still, a LR(1) parser (like `wesl-rs`) can parse this, with the limitation above.

stefnotch:
I'm happy with it! My only concern is that if we proposed it to the committee, they'd have a reason for saying *nono* and point at "it breaks LALR or LR(1) parsers".

But that's a problem for future us :D So feel free to merge and ship it!

PR #172, opened 2025-12-13
Update WeslToml.md with more precise algorithm for include/exclude

stefnotch:
I noticed that the include/exclude section doesn't properly specify what to do in certain edge cases.

1. What to do when there's a nested `wesl.toml` file
2. If I have `root = "./shaders"` combined with `include = ["foo/bar"]`, then what should happen? Should I still look at `foo/bar`, or should it be ignored since it's nonsense

### DIFF
--- a/WeslToml.md
+++ b/WeslToml.md
@@ -92,6 +92,11 @@ The patterns are glob patterns, which support
 
 If the last path segment does not contain a file extension or wildcard, then it is treated as a directory, and files with `.wesl` or `.wgsl` extensions inside that directory are included.
 
+To implement this, one starts at the `root` folder, and recursively goes over the children.
+1. If a `wesl.toml` file is present, that entire subtree is excluded, as it is under the purview of another `wesl.toml` file.
+2. If the file/folder path matches any exclusion, that entire subtree is excluded.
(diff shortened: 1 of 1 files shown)


stefnotch:
This doesn't cover the "implicit wesl.toml" case yet. As in, whenever we see a package.json, we should pretend that there's a wesl.toml next to it.

mighdoll:
Good question on the nested case. If it's not clear what to do, could defer if it's not needed by any current projects. 

another future case to think about. with include/exclude alone, exclude takes priority. if we decide to respect gitignore, is the priority: include < gitignore < exclude?

re: nonsensical mixes of root and include, should we warn?

mighdoll:
re: nonsensical `root` and `include`, how 'bout warn and drop the `root`? Seems like a common scenario given our defaults. Users add a `version` and an `include`, but don't add a `root`.

stefnotch:
It doesn't make too much sense when you look at that repo. It becomes more obvious once you look at Bevy itself, and realize that every "crate" is an independently published project https://github.com/bevyengine/bevy/tree/main/crates

It's possible to only use bevy_render, and supply the rest of the renderer yourself. (I think)

At the same time, they want users that only depend on Bevy to only have import statements like `import bevy::... `

stefnotch:
Closing in favour of #182

PR #173, opened 2025-12-19
Revise Packaging.md

k2d222:
It's time to un-draft Packaging.md

### DIFF
--- a/Packaging.md
+++ b/Packaging.md
@@ -1,4 +1,4 @@
-.# Packaging
+# Packaging
 
 WESL enables shader packages for reusing shader code by other packages or applications. Shader packages are published to repositories such as [npm] or [crates.io].
 
--- a/Packaging.md
+++ b/Packaging.md
@@ -1,4 +1,4 @@
-# Packaging
+.# Packaging
 
 WESL enables shader packages for reusing shader code by other packages or applications. Shader packages are published to repositories such as [npm] or [crates.io].
 
@@ -30,7 +30,7 @@ You can publish your package with whatever name you like. We however recommend f
 
 (TODO: unification with param const?)
 
-If two packages in the dependency tree are [semver-compatible](https://semver.org/), the WESL linker will unify them, meaning it will include only one version of the package in its output (usually the highest semver-compatible version available).
+If two packages in the dependency tree are [semver-compatible](https://semver.org/), npm or Cargo will likely unify them, meaning it will include only one version of the package in its output (often the highest semver-compatible version available).
 This unification can have observable side-effects that a user must be warry of. Module-scope declarations may, or may not be duplicated.
 
 ### Example
@@ -49,7 +49,7 @@ fn rand() -> f32 {
 
 Here, the `random::rand()` function has internal state represented by `prng_state`.
 If two packages depend on semver-compatible versions of `random`, then they would share that internal state.
-If however they depend on non-semver-compatible versions, two copies of the state and the `rand()` function will be used, suddenly producing different results.
+If however they depend on non-semver-compatible versions, two copies of the state and the `rand()` function can be used, suddenly producing different results.
 
 ## Visibility
 
--- a/Packaging.md
+++ b/Packaging.md
@@ -4,17 +4,18 @@ WESL enables shader packages for reusing shader code by other packages or applic
 
 ## Using shader packages
 
-Dependencies for your application are defined in the [`wesl.toml`] file. The WESL linker is responsible for finding and downloading dependencies: [wesl-js] will fetch packages from [npm] and [wesl-rs] will fetch them from [crates.io].
+Dependencies for your application are defined in the [`wesl.toml`] file.
+The WESL linker is responsible for finding and downloading dependencies: [wesl-js] will fetch packages from [npm], while [wesl-rs] will fetch them from [crates.io].
 Inside your shader modules, you can reach declarations in dependencies with import paths, as explained in [imports].
 
 ## Creating and publishing shader packages
 
-Any [`wesl.toml`] file declares a new shader package, which can be published by the linker's CLI.
+Any [`wesl.toml`] file declares a new shader package, which can be published using the linker's CLI.
 Publishing packages is implementation-specific, follow the instructions for your linker ([wesl-js] or [wesl-rs]).
 
 ### Package naming guidelines
 
-You can publish your package with whatever name you like. We however recommend following these recommendations so your package can be easily found by searching the package registry.
+You can publish your package with whatever name you like. We however recommend following these guidelines so your package can be easily found and consumed by end users.
 
 * Use a name that is also a valid WGSL identifier. Otherwise, end-users will have to rename it in [`wesl.toml`].
 * Add the `_wgsl` suffix to the name: `mypackage_wgsl`.
--- a/Packaging.md
+++ b/Packaging.md
@@ -1,57 +1,67 @@
 # Packaging
-(TBD)
 
-The section will discuss packaging WESL in reusable crates or npm packages.
+WESL enables shader packages for reusing shader code by other packages or applications. Shader packages are published to repositories such as [npm] or [crates.io].
 
-See also [Visibility](Visibility.md).
+## Using shader packages
 
-* What goes in `package.json`?
-  * at runtime in the browser, the linker needs
-    WESL strings and relative paths for every needed WESL file, from every every package.
-    * (this so the linker can resolve imports)
-  * something needs to tell javascript bundlers to include needed WESL files in the bundle:
-    * we could ask package publishers to produce a javascript file containing a map of relative paths and WESL strings
-    * perhaps better would be to have a vite/rollup plugin trace from the source WESL to the needed package WESL files (using the linker as a library to parse WESL)
-      * this would avoid requiring publishers to manually publish string maps of WESL text,
-        instead publishers would just publish files in dist.
-        * the consumer plugin tool would bundle files into strings.
-      * corner case issue: runtime conditional compilation could potentially change the imported
-        files required. Perhaps we'll need some workaround markers in this rare case, so tree shaking can still work.
-    * note that the WESL-linker already has this problem of bundling local WGSL files
-      and asks users to 'self package' WGSL using import.meta.glob
-      (which bundles files into [file path, WGSL string] pairs)
-  * what other metadata should be available to the linker?
-    * name of package.
-    * host visible bits from WGSL? e.g. entry points, runtime variables, overrides, binding groups, uniforms?
-      An IDE tool could use that data to typecheck/autocomplete host calls.
-      Things like entry points are computable from the WGSL,
-      but it's a lot to ask of a language server to parse through all the WGSL in advance to find them.
-      Probably better to put it into the metadata if we need it.
-      * these are parts of the API of the package as a whole and can reduce bugs by making them visible
-  * how do we handle packages with multiple entry points e.g. stoneberry/reduce stoneberry/prefixSum.
-* What goes in `cargo.toml`?
+Dependencies for your application are defined in the [`wesl.toml`] file. The WESL linker is responsible for finding and downloading dependencies: [wesl-js] will fetch packages from [npm] and [wesl-rs] will fetch them from [crates.io].
+Inside your shader modules, you can reach declarations in dependencies with import paths, as explained in [imports].
 
-## `wesl.toml` file
-The `wesl.toml` file is a configuration file that tells the language server where to find the packages. It is similar to a `Cargo.toml` file in that regard.
+## Creating and publishing shader packages
 
-```toml
-name = "my"
-edition = "2024"
+Any [`wesl.toml`] file declares a new shader package, which can be published by the linker's CLI.
+Publishing packages is implementation-specific, follow the instructions for your linker ([wesl-js] or [wesl-rs]).
 
-[dependencies]
-bevy_ui = { cargo = "bevy_ui" }
-shader_wiz = { path = "./node_modules/shader_wiz/src/main.wgsl" }
+### Package naming guidelines
+
+You can publish your package with whatever name you like. We however recommend following these recommendations so your package can be easily found by searching the package registry.
+
+* Use a name that is also a valid WGSL identifier. Otherwise, end-users will have to rename it in [`wesl.toml`].
+* Add the `_wgsl` suffix to the name: `mypackage_wgsl`.
+* Use snake_case (underscores to separate words): `my_great_package_wgsl`.
+* If your package is part of a larger project, or produced by a company, you can prefix it with that name: `mycompany_mypackage_wgsl`.
+  * For [wesl-js] specifically, you can use the common naming convention `@mycompany/mypackage_wgsl`
+* Look for packages with the same name in both [npm] and [crates.io].
+  * It is courtesy to leave the name free for the original author if they wish to publish to the other registry.
+  * It also avoids confusion for end-users who may think it is the same package.
+
+## Semver-compatibility and dependency unification
+
+(TODO: unification with param const?)
+
+If two packages in the dependency tree are [semver-compatible](https://semver.org/), the WESL linker will unify them, meaning it will include only one version of the package in its output (usually the highest semver-compatible version available).
+This unification can have observable side-effects that a user must be warry of. Module-scope declarations may, or may not be duplicated.
+
+### Example
+
+```wgsl
+// package random
+// --------------
+var<private> prng_state: f32 = 0;
+
+// return a random float in [0, 1]. Do not actually use this function, it is awful.
+fn rand() -> f32 {
+    prng_state = fract(sin(x)*12345.6789);
+    return prng_state;
+}
 ```
 
-We specify how to resolve the paths of packages instead of scanning folders for `*.wgsl` files.
-In the Javascript world, it is common to have a `node_modules` folder with 10k files, which is not practical for a language server to scan.
+Here, the `random::rand()` function has internal state represented by `prng_state`.
+If two packages depend on semver-compatible versions of `random`, then they would share that internal state.
+If however they depend on non-semver-compatible versions, two copies of the state and the `rand()` function will be used, suddenly producing different results.
 
-We are planning on taking advantage of existing package managers, such as `cargo` for Rust, and `npm` for Javascript. This makes it easier for users to consume shaders, and makes sense for ecosystem-specific tools.
+## Visibility
 
-### Unresolved questions
-* Is .toml the best file format for the `wesl.toml`? Some alternatives would be JSON/JSON5 and StrictYAML.
-* do we need to put paths for every package's WGSL in `wesl.toml`?
-  * Fine to get started that way if need be, but its a maintenance burden every user..
-  * Better if we could get the language servers to find the WGSL in packages in node_modules.
-* Do we need to list the WGSL package names in `wesl.toml`?
-  They'll already be listed in as `cargo.toml` / `package.json`..
+All declarations reachable from a direct dependency's root module can be accessed from the parent package.
+Declarations in indirect dependencies (i.e., dependencies of dependencies) are never reachable.
+
+Future iterations of WESL may introduce a re-export and/or visibility control mechanism. See [visibility].
+
+
+[npm]: https://www.npmjs.com/
+[crates.io]: https://crates.io/
+[`wesl.toml`]: WeslToml.md
+[imports]: Imports.md
+[visibility]: Visibility.md
+[wesl-js]: https://github.com/wgsl-tooling-wg/wesl-js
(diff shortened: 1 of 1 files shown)


Packaging.md
```diff
@@ -1,57 +1,68 @@
...
+  * For [wesl-js] specifically, you can use the common naming convention `@mycompany/mypackage_wgsl`
+* Look for packages with the same name in both [npm] and [crates.io].
+  * It is courtesy to leave the name free for the original author if they wish to publish to the other registry.
+  * It also avoids confusion for end-users who may think it is the same package.
+
+## Semver-compatibility and dependency unification
+
+(TODO: unification with param const?)
+
+If two packages in the dependency tree are [semver-compatible](https://semver.org/), the WESL linker will unify them, meaning it will include only one version of the package in its output (usually the highest semver-compatible version available).
```
stefnotch: Will the linker do this, or will the package manager already do it?

Packaging.md
```diff
@@ -1,57 +1,68 @@
...
+This unification can have observable side-effects that a user must be warry of. Module-scope declarations may, or may not be duplicated.
```
stefnotch: *wary

Packaging.md
```diff
@@ -1,57 +1,68 @@
...
+You can publish your package with whatever name you like. We however recommend following these guidelines so your package can be easily found and consumed by end users.
+
+* Use a name that is also a valid WGSL identifier. Otherwise, end-users will have to rename it in [`wesl.toml`].
+* Add the `_wgsl` suffix to the name: `mypackage_wgsl`.
+* Use snake_case (underscores to separate words): `my_great_package_wgsl`.
+* If your package is part of a larger project, or produced by a company, you can prefix it with that name: `mycompany_mypackage_wgsl`.
+  * For [wesl-js] specifically, you can use the common naming convention `@mycompany/mypackage_wgsl`
+* Look for packages with the same name in both [npm] and [crates.io].
+  * It is courtesy to leave the name free for the original author if they wish to publish to the other registry.
+  * It also avoids confusion for end-users who may think it is the same package.
```
stefnotch, line 47: I like this :)

Packaging.md
k2d222: I guess the package mgr should, although idk how that works in js currently, and in rs this is not the current behavior. Currently, it will unify versions, but still duplicate contents for all subdependencies. I wonder if we should actually stabilize this behavior or just tell users to pay attention.

stefnotch: I think we should just tell users to pay attention, because package managers can and will unify dependencies.

A more thorough solution is going to require a whole lot more investigating. I bet there are a few edge cases that we aren't even aware of at the moment.

Packaging.md
```diff
@@ -1,57 +1,68 @@
...
-
-[dependencies]
-bevy_ui = { cargo = "bevy_ui" }
-shader_wiz = { path = "./node_modules/shader_wiz/src/main.wgsl" }
+
+WESL enables shader packages for reusing shader code by other packages or applications. Shader packages are published to repositories such as [npm] or [crates.io].
+
+## Using shader packages
+
+Dependencies for your application are defined in the [`wesl.toml`] file.
```
mighdoll: dependencies in wesl.toml are optional for the wesl-js package (it gets them from package.json)

Packaging.md
```diff
@@ -1,57 +1,68 @@
...
+* Use snake_case (underscores to separate words): `my_great_package_wgsl`.
+* If your package is part of a larger project, or produced by a company, you can prefix it with that name: `mycompany_mypackage_wgsl`.
+  * For [wesl-js] specifically, you can use the common naming convention `@mycompany/mypackage_wgsl`
+* Look for packages with the same name in both [npm] and [crates.io].
+  * It is courtesy to leave the name free for the original author if they wish to publish to the other registry.
+  * It also avoids confusion for end-users who may think it is the same package.
+
+## Semver-compatibility and dependency unification
+
+(TODO: unification with param const?)
```
mighdoll: suggest we remove TODO so this doesn't look incomplete. When param const comes in (soon I hope!) we can revise as appropriate.

Packaging.md
```diff
@@ -1,57 +1,68 @@
...
+
+If two packages in the dependency tree are [semver-compatible](https://semver.org/), npm or Cargo will likely unify them, meaning it will include only one version of the package in its output (often the highest semver-compatible version available).
```
mighdoll: I think we might say that users are doing this unification. i.e. users of npm and cargo may unify them.

Packaging.md
```diff
@@ -1,57 +1,68 @@
...
+This unification can have observable side-effects that a user must be wary of. Module-scope declarations may, or may not be duplicated.
```
mighdoll: I think "This unification can have osbservable side-effects" is enough. (no need to scare people/agents away from using packages)

Packaging.md
```diff
@@ -1,57 +1,68 @@
...
+    prng_state = fract(sin(x)*12345.6789);
+    return prng_state;
+}
 ```
 
-We specify how to resolve the paths of packages instead of scanning folders for `*.wgsl` files.
-In the Javascript world, it is common to have a `node_modules` folder with 10k files, which is not practical for a language server to scan.
+Here, the `random::rand()` function has internal state represented by `prng_state`.
+If two packages depend on semver-compatible versions of `random`, then they would share that internal state.
+If however they depend on non-semver-compatible versions, two copies of the state and the `rand()` function can be used, suddenly producing different results.
```
mighdoll: s/suddenly//

Packaging.md
```diff
@@ -1,57 +1,68 @@
...
+
+## Visibility
+
+All declarations reachable from a direct dependency's root module can be accessed from the parent package.
+Declarations in indirect dependencies (i.e., dependencies of dependencies) are never reachable.
```
mighdoll: 'reachable' is a little confusing here. The declarations can of course be used indirectly. and does the root module matter here? 
I think the point is that indirect package dependencies (via wesl.toml or package.json or cargo.toml) are not usable in WESL statements. You can only say `import foo_pkg::bar;` if `foo_pkg` is a direct dependency.

Packaging.md
mighdoll: Oh, I see, the point is that package managers will sometimes unify by default. I think that's the default for npm, though the rules are overridable and expectations vary on whether indirect dependencies are unified. (AFAIR by default pnpm doesn't unify indirect dependencies, but npm might still do that)

Packaging.md
```diff
@@ -1,57 +1,68 @@
...
+
+### Package naming guidelines
+
+You can publish your package with whatever name you like. We however recommend following these guidelines so your package can be easily found and consumed by end users.
+
+* Use a name that is also a valid WGSL identifier. Otherwise, end-users will have to rename it in [`wesl.toml`].
+* Add the `_wgsl` suffix to the name: `mypackage_wgsl`.
+* Use snake_case (underscores to separate words): `my_great_package_wgsl`.
+* If your package is part of a larger project, or produced by a company, you can prefix it with that name: `mycompany_mypackage_wgsl`.
+  * For [wesl-js] specifically, you can use the common naming convention `@mycompany/mypackage_wgsl`
```
mighdoll: we might add a note here that package names with @ and - and / are sanitized for use in WESL:
"@scope/my-package" => `import scope__my_package::...`

mighdoll (commented):
It'll be good to get this in! sorry for forgetting about it and glad you're pushing it forward. lmk if you want help on any of the minor edits, I can make more specific text proposals if needed.

Packaging.md
k2d222: lemme know if the current revision clarifies.

Packaging.md
k2d222: removed mentions of "reachable" and "root module". Instead I phrased as "can be imported"

Packaging.md
k2d222: I don't know the exact sanitization rules in -js, so I added vague footnote

Packaging.md
```diff
@@ -1,57 +1,71 @@
...
-
-```toml
-name = "my"
-edition = "2024"
-
-[dependencies]
-bevy_ui = { cargo = "bevy_ui" }
-shader_wiz = { path = "./node_modules/shader_wiz/src/main.wgsl" }
+
+WESL enables shader packages for reusing shader code by other packages or applications. Shader packages are published to repositories such as [npm] or [crates.io].
```
mighdoll: perhaps:

shader packager are *typically* published

(still a package if it's not published)

Packaging.md
```diff
@@ -1,57 +1,71 @@
...
+
+## Using shader packages
+
+The WESL linker is responsible for finding and downloading dependencies: [wesl-js] will fetch packages from [npm], while [wesl-rs] will fetch them from [crates.io].
```
mighdoll: Something like this perhaps:

In development, packages are typically loaded from repositories by ecosystem package managers like npm or cargo. 

(Tools like wgsl-play can fetch packages directly from the web.  'typically' leaves room for that w/o getting into those variations.)

Packaging.md
```diff
@@ -1,57 +1,71 @@
...
+The package dependencies of a shader project are either discovered automatically by the WESL linker, or explicitly listed in the `dependencies` section of [`wesl.toml`].
```
mighdoll: wesl-rs currently uses build.rs? 
wesl-js normally infers from shader code + package.json.
(I don't think we actually use wesl.toml deps in linking at present..)

Packaging.md
```diff
@@ -1,57 +1,71 @@
...
+Inside shader modules, one can reach declarations in dependencies with import paths, as explained in [imports].
```
mighdoll: perhaps something in this direction:

WESL tools resolve module paths in WESL code using the algorithm described in [import resolution](Imports.md#import-resolution-algorithm). Each ecosystem package format provides a standard way for WESL tools to map from module paths to shader code.

Packaging.md
```diff
@@ -1,57 +1,71 @@
...
+
+## Creating and publishing shader packages
+
+Any [`wesl.toml`] file declares a new shader package, which can be published using the linker's CLI.
```
mighdoll, line 10: in js, [wesl-packager](https://www.npmjs.com/package/wesl-packager) is currently a separate cli (from the wesl-link cli).

Packaging.md
```diff
@@ -1,57 +1,71 @@
...
+Publishing packages is implementation-specific, linkers ([wesl-js] or [wesl-rs]) provide documentation on how to publish to their respective package repositories.
+
+### Package naming guidelines
+
+A package can be published under any name. We however recommend following these guidelines, so the package can be easily found and consumed by end users.
+
+* Use a name that is also a valid WGSL identifier. Otherwise, end-users will have to rename it in [`wesl.toml`].
```
mighdoll, line 40: (I think we're leaning towards renaming in package.json/cargo.toml, but it hasn't settled yet)

Packaging.md
```diff
@@ -1,57 +1,71 @@
...
+* Add the `_wgsl` suffix to the name: `mypackage_wgsl`.
+* Use snake_case (underscores to separate words): `my_great_package_wgsl`.
+* If your package is part of a larger project, or produced by a company, you can prefix it with that name: `mycompany_mypackage_wgsl`.
+  * For [wesl-js] specifically, you can use the common naming convention `@mycompany/mypackage_wgsl`.[^1]
+* Look for packages with the same name in both [npm] and [crates.io].
+  * It is courtesy to leave the name free for the original author if they wish to publish to the other registry.
+  * It also avoids confusion for end-users who may think it is the same package.
+
+[^1]: When a package name contains `/` or `@`, [wesl-js] will sanitize it to become a valid WGSL identifier. Refer to its documentation for more information.
```
mighdoll: Also '-' (which also wesl-rs sanitizes?).

And an example here would be useful:
`@mycompany/mypackage_wgsl` is imported as `mycompany__mypackage_wgsl::…`

Packaging.md
```diff
@@ -1,57 +1,71 @@
...
+fn rand() -> f32 {
+    prng_state = fract(sin(x)*12345.6789);
+    return prng_state;
+}
 ```
 
-We specify how to resolve the paths of packages instead of scanning folders for `*.wgsl` files.
-In the Javascript world, it is common to have a `node_modules` folder with 10k files, which is not practical for a language server to scan.
+In this example, the `random::rand()` function has internal state represented by `prng_state`.
+If two packages depend on semver-compatible versions of `random`, there will be two copies the `rand()` function and its internal state. Calls to `rand()` from both packages refer to the same declaratione.
```
mighdoll: two copies the => a single copy of the 

declaratione => declaration

Packaging.md
```diff
@@ -1,57 +1,71 @@
...
+If however they depend on non-semver-compatible versions, there will be two copies the `rand()` function and its internal state. Calls to `rand()` from both packages refer to distinct declarations.
+
+## Visibility
```
mighdoll: Suggest renaming to e.g., Package Visibility, or Dependency Visibility
to distinguish from public/private declarations stuff in visibility.md

Packaging.md
```diff
@@ -1,57 +1,71 @@
...
+
+A package can only import declarations in its direct dependencies, using an import statement or an inline import (see [imports]).
```
mighdoll: perhaps:

A package can only import from packages that are direct dependencies...

(dropping 'declarations' within the package which amply discussed in visibility.md, to which you already refer)

Packaging.md
```diff
@@ -1,57 +1,71 @@
...
+If two packages depend on semver-compatible versions of `random`, there will be two copies the `rand()` function and its internal state. Calls to `rand()` from both packages refer to the same declaratione.
+If however they depend on non-semver-compatible versions, there will be two copies the `rand()` function and its internal state. Calls to `rand()` from both packages refer to distinct declarations.
+
+## Visibility
+
+A package can only import declarations in its direct dependencies, using an import statement or an inline import (see [imports]). 
+It cannot import declarations in its indirect dependencies (i.e., dependencies of dependencies).
+Declarations in indirect dependencies can be indirectly used by the package, for instance if a direct dependency provides a function which calls a function in one of its dependencies.
+
+Future iterations of WESL may introduce a re-export and/or visibility control mechanism. See [visibility].
```
mighdoll: Drop the "future" framing?  the current visibility doc marks itself as future, so we don't actually have to mark it here.. (then the text doesn't need to change when visibility lands.)

mighdoll (changes requested):
Looks good, nice work moving this forward! Suggestions below are mostly editorial.

Packaging.md
```diff
@@ -1,57 +1,71 @@
...
+    prng_state = fract(sin(x)*12345.6789);
+    return prng_state;
+}
 ```
 
-We specify how to resolve the paths of packages instead of scanning folders for `*.wgsl` files.
-In the Javascript world, it is common to have a `node_modules` folder with 10k files, which is not practical for a language server to scan.
+In this example, the `random::rand()` function has internal state represented by `prng_state`.
+If two packages depend on semver-compatible versions of `random`, there will be two copies the `rand()` function and its internal state. Calls to `rand()` from both packages refer to the same declaratione.
+If however they depend on non-semver-compatible versions, there will be two copies the `rand()` function and its internal state. Calls to `rand()` from both packages refer to distinct declarations.
```
mighdoll: two copies the => two copies of the

Packaging.md
k2d222: > wesl-rs currently uses build.rs?

yes, packages are built in the build script. Currently the -rs expects that packages are manually provided in the build command. 

added a note to be removed once we have better sopport of wesl-toml.

Packaging.md
k2d222: my goal was to explain how packages can be used inside shader code, and that is using import statment. I've settled on a sentence in-between yours and mine.

Packaging.md
```diff
@@ -1,57 +1,70 @@
...
+
+* Use a name that is also a valid WGSL identifier. Otherwise, end-users will have to rename it in [`wesl.toml`].
+* Add the `_wgsl` suffix to the name: `mypackage_wgsl`.
+* Use snake_case (underscores to separate words): `my_great_package_wgsl`.
+* If your package is part of a larger project, or produced by a company, you can prefix it with that name: `mycompany_mypackage_wgsl`.
+  * For [wesl-js] specifically, you can use the common naming convention `@mycompany/mypackage_wgsl`.[^2]
+* Look for packages with the same name in both [npm] and [crates.io].
+  * It is courtesy to leave the name free for the original author if they wish to publish to the other registry.
+  * It also avoids confusion for end-users who may think it is the same package.
+  *
```
experimentslee: extra bullet

Packaging.md
```diff
@@ -1,57 +1,70 @@
...
-bevy_ui = { cargo = "bevy_ui" }
-shader_wiz = { path = "./node_modules/shader_wiz/src/main.wgsl" }
+
+WESL enables shader packages for reusing shader code by other packages or applications. Shader packages are typically published to repositories such as [npm] or [crates.io].
+
+## Using shader packages
+
+In development, packages are typically loaded from repositories by ecosystem package managers like npm or cargo. [wesl-js] will fetch packages from [npm], while [wesl-rs] will fetch them from [crates.io].
+The package dependencies of a shader project are either provided manually using linker-specific settings, discovered automatically by the linker, or explicitly listed in the `dependencies` section of [`wesl.toml`][^1].
+A package contents is reachable from user code via module paths starting with the package name. This mechanism is detailed in the [import resolution algorithm](Imports.md#resolving-a-declaration-path).
```
experimentslee: package's contents are

Packaging.md
```diff
@@ -1,57 +1,70 @@
...
+[npm]: https://www.npmjs.com/
+[crates.io]: https://crates.io/
+[`wesl.toml`]: WeslToml.md
+[imports]: Imports.md
+[visibility]: Visibility.md
+[wesl-js]: https://github.com/wgsl-tooling-wg/wesl-js
+[wesl-rs]: https://github.com/wgsl-tooling-wg/wesl-rs
 
-We are planning on taking advantage of existing package managers, such as `cargo` for Rust, and `npm` for Javascript. This makes it easier for users to consume shaders, and makes sense for ecosystem-specific tools.
+[^1]: At the time of writing, [wesl-js] and [wesl-rs] do not support reading dependencies in `wesl.toml`, but that will become the preferred way of discovering dependencies in the near future.
```
experimentslee: perhaps just drop the footnote for now, the base sentence covers the options already we don't need to predict in a spec footnote that will soon go out of date.

(also, I think wesl-js probably stays default auto, and maybe still ignores the wesl.toml deps? not sure, but we'll discuss with #194)

Packaging.md
```diff
@@ -1,57 +1,70 @@
...
+[^2]: When a package name contains symbols such as `-` `/` or `@`, [wesl-js] will sanitize it to become a valid WGSL identifier, e.g. a package named `@mycompany/mypackage_wgsl` can be imported with the prefix `mycompany__mypackage_wgsl`. [wesl-js] forbids symbols, with the exception of `-` (dash) which is replaced with `_` (underscore). Refer to the linkers' documentation for more information.
```
experimentslee: for wesl-js, it's: 
- `@` is removed
- `/` becomes `__`
- `-` becomes `_`

I think maybe the second ref has a typo, supposed to be to wesl-rs.

I think it's goodness to mention the sanitizations, since we have to keep them stable.

experimentslee (changes requested):
looking good. best not promise to wesl.toml deps for wesl-js. 
resolve/review the 4 comments and then LGTM w/o further review

Packaging.md
k2d222: instead of a footnote, I'll write a subsection "package name sanitization". It deserves q deeper explanation.

stefnotch (approved):

Packaging.md
```diff
@@ -2,30 +2,50 @@
...
 
+### Accessing packages from shader code
+
+Packages' contents are reachable from user code via module paths prefixed with the package name, e.g. `package_name::declaration_name`. This mechanism is detailed in the [import resolution algorithm](Imports.md#resolving-a-declaration-path).
+The package name in shader code is the same as the name published on [npm] or [crates.io], with two exceptions:
+
+1. The name is sanitized to remove certain common symbols that are invalid in WGSL identifiers (see following section).
+2. The name can be overriden in the dependencies list in [`wesl.toml`], e.g. `wgsl_name = { package = "@published/name" }`.
+
+If after these operations, the shader name is still not a valid WGSL identifier, the package will not be reachable from user code.
```
k2d222, line 21: I wonder if it should be an error instead. i.e. the linker must check that the package name is a valid WGSL ident. Does -js do that? @mighdoll

stefnotch (approved):

Packaging.md
mighdoll: The packager tool should error and it doesn't. Good one. (the -js library consumer would fail because it's unreachable. The consumer could workaround with a package.json rename, but that's a bit late)

PR #182, opened 2026-04-19
Fixes to wesl.toml

### LINKED ISSUE #175: package keys to root of wesl.toml
A simple app or library would have a wesl.toml that looks something like this:

```toml
[package]
edition = "2026_pre"
```

I'd be nice to make the `[package]` part unnecessary. That'd be one less thing to learn for the simple case. 

We can simplify things for users by putting the package keys at the root of the toml. bun chose this approach for their toml config, see: [bunfig.toml](https://bun.com/docs/runtime/bunfig#runtime). Like us, bun expects their config to be a semi-optional adjunct to the required native config (package.json, cargo.toml).

I tried a PR for wesl-rs that parses the wesl.toml and makes the `[package]` optional.  In that rev, the package keys work both at the root of toml or under `[package]`. So we could support it both ways if we'd like.

### LINKED ISSUE #178: Update wesl.toml spec
We should
- [ ] Settle on some edition names. mighdoll suggested 2025-1 instead of the `unstable_2025`
- [ ] Remove the `[package]`
- [ ] Document that `include` may only include files and directories from `root = "some_dir"`
- [ ] Document an especially cursed case with nested `wesl.toml`s

### Especially cursed case

```
wesl.toml with root = "./shaders/foo"
/shaders
  wesl.toml with root = "./bar"
  /foo
    main.wesl
  /bar
    other.wesl
```

When the IDE looks at `main.wesl`, it has to go up the directory tree to find the corresponding `wesl.toml`. There, it would have to skip the first one it finds (`wesl.toml with root = "./bar"`) and go to the topmost one (`wesl.toml with root = "./shaders/foo"`), because of how the roots are set up.

One can construct even more cursed cases with `include`s. Like
```
wesl.toml with root = "." and include = ["./**/test*.wesl"]
/shaders
  wesl.toml with root = "." and exclude = ["./**/test*.wesl"]
```

Here two files that are next to each other can be inside of a completely different `package::`. (one says `include` and the other one says `exclude`)

While we could support all that, @mighdoll suggested that we just don't. Easier to just disallow it. The first parent `wesl.toml` is the corresponding one and if it doesn't match, then we have a standalone file.

stefnotch:
Fix #178
Fix #175

Supersedes https://github.com/wgsl-tooling-wg/wesl-spec/pull/172

### DIFF
--- a/WeslToml.md
+++ b/WeslToml.md
@@ -5,7 +5,6 @@ WESL shader projects have a standardized `wesl.toml` file, which describes the p
 The format looks as follows. As much of it is optional as possible.
 
 ```toml
-[package]
 # Version of WESL used in this project.
 edition = "2026_pre"
 
@@ -46,6 +45,10 @@ As your application grows, we suggest adding a `wesl.toml`. This lets you pin do
 
 Specifies which edition is used by wesl. All wesl editions can be used.
 
+Supported editions are
+- `WGSL` for pure WGSL
+- `2026_pre` for the current WESL version
+
 - Mandatory
   - If `wesl.toml` is not present, the default is the latest edition.
   - WESL tools will provide long-term support for stable LTS editions, ensuring forwards compatibility.
@@ -84,12 +87,18 @@ Specifies an array of patterns where wesl files are located. The patterns are re
 The default value of `include` is all wesl and wgsl files in the `root` directory, recursively.
 The default value of `exclude` is an empty array.
 
-The patterns are glob patterns, which support
+The patterns are glob patterns, made up of the following components
+- file and folder names
 - `*` for zero or more characters in file/folder names
 - `?` for one character in a file/folder name
 - `**/` for any directory nested to any level
 
 If the last path segment does not contain a file extension or wildcard, then it is treated as a directory, and files with `.wesl` or `.wgsl` extensions inside that directory are included.
+
+To implement this, one starts at the `root` folder, and recursively goes over the children.
+1. If a `wesl.toml` file is present, that entire subtree is excluded, as it is under the purview of another `wesl.toml` file.
+2. If the file/folder path matches any exclusion, that entire subtree is excluded.
+3. If the file/folder path does *not* match the *prefix* of any of the inclusions, then that subtree is excluded.
 
 ### Path semantics
 
--- a/WeslToml.md
+++ b/WeslToml.md
@@ -46,7 +46,6 @@ As your application grows, we suggest adding a `wesl.toml`. This lets you pin do
 Specifies which edition is used by wesl. All wesl editions can be used.
 
 Supported editions are
-- `WGSL` for pure WGSL
 - `2026_pre` for the current WESL version
 
 - Mandatory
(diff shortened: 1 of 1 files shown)


stefnotch:
Can we merge this? It's affecting a user

https://github.com/webgpu-tools/wesl-spec/issues/175

stefnotch:
@mighdoll @k2d222 Polite ping, requesting a review :)

k2d222 (approved):

WeslToml.md
```diff
@@ -46,6 +45,10 @@ As your application grows, we suggest adding a `wesl.toml`. This lets you pin do
 
 Specifies which edition is used by wesl. All wesl editions can be used.
 
+Supported editions are
+- `WGSL` for pure WGSL
```
mighdoll: I could imagine 'WGSL' makes sense for some situations, but I think we should spell out the semantics if we're going to add this.  Seems potentially user confusing: since we're moving away from public by default a pure WGSL package can't mark anything `public`, so it exports nothing and isn't usable as a library. Imports between files don't work either (WGSL has no import statements).

(Or we could drop the 'WGSL' edition if the need isn't clear yet.)

WeslToml.md
```diff
@@ -84,13 +87,19 @@ Specifies an array of patterns where wesl files are located. The patterns are re
 The default value of `include` is all wesl and wgsl files in the `root` directory, recursively.
 The default value of `exclude` is an empty array.
 
-The patterns are glob patterns, which support
+The patterns are glob patterns, made up of the following components
+- file and folder names
 - `*` for zero or more characters in file/folder names
 - `?` for one character in a file/folder name
 - `**/` for any directory nested to any level
```
mighdoll: Clarify that `**` matches zero or more segments, so `shaders/**/*.wesl` also includes `shaders/top.wesl` (a file directly in `shaders`, not just nested ones).

WeslToml.md
```diff
@@ -84,13 +87,19 @@ Specifies an array of patterns where wesl files are located. The patterns are re
...
 
 If the last path segment does not contain a file extension or wildcard, then it is treated as a directory, and files with `.wesl` or `.wgsl` extensions inside that directory are included.
 
+To implement this, one starts at the `root` folder, and recursively goes over the children.
+1. If a `wesl.toml` file is present, that entire subtree is excluded, as it is under the purview of another `wesl.toml` file.
```
mighdoll: Sounds like this would exclude the root wesl.toml..

WeslToml.md
```diff
@@ -84,13 +87,19 @@ Specifies an array of patterns where wesl files are located. The patterns are re
...
+2. If the file/folder path matches any exclusion, that entire subtree is excluded.
+3. If the file/folder path does *not* match the *prefix* of any of the inclusions, then that subtree is excluded.
```
mighdoll: What about globs? Prefix-matching against a glob like `shaders/**/*.wesl` is ambiguous.

WeslToml.md
```diff
@@ -84,13 +87,19 @@ Specifies an array of patterns where wesl files are located. The patterns are re
...
+
```
mighdoll, line 104: Perhaps something like this as an alternative:

A file under root is included when all of these hold:

1. it is not inside a subtree owned by a nested wesl.toml (the current package's own wesl.toml does not count);
2. its path, relative to this wesl.toml, matches at least one include glob;
3. its path matches no exclude glob.

(non-normative) Implementations may prune directories while walking: skip a folder if it contains a nested wesl.toml, if it matches an exclude, or if no include glob could match any path beneath it.

mighdoll (changes requested):
nested wesl.toml sounds good, some suggestions inline on spelling that in the spec.

WeslToml.md
stefnotch: I'm removing the WGSL edition, it doesn't have a real purpose

WeslToml.md
stefnotch: I changed the phrasing, hopefully it's clearer now

WeslToml.md
stefnotch: Good catch, will fix

WeslToml.md
stefnotch: I like this a lot more, thank you!

WeslToml.md
stefnotch: fixed by copying your homework :P

stefnotch:
Okay, comments are addressed. Ready for merge?

k2d222 (approved):

mighdoll (approved):
lgtm!

Issue #130, opened 2025-05-12
Support import package::*; to import all items in a file


newDINO:
Current implementation and spec doesn't seem to support `import package::*;`

mighdoll:
'tis true. We worry that star imports would lead to maintenance problems.. e.g. if someone adds `foo` to a module that you're star importing, and that creates a warning because you have a local `fn foo`.  But maybe that concern is overblown. Have you already run into this in practice with shaders?

newDINO:
Not really. I don't have that many imports from one file yet. But star importing would indeed be more convenient in certain cases. But it would increase the complexity of wesl and may not be that useful since shaders are usually simple programs and writing too complex shaders will increase the pressure of GPU registers resulting in poor performance.

I used to have a system base on `include` to handle shader reusing. When switching to wesl, the easiest method is using star importing. But now I have completed the migration, I find that I don't actually use that many imports.

k2d222:
coming back to this issue with a use-case.
"prelude" libraries can be quite useful to extend or override the set of builtin declarations. Importing all used stuff can be tedious and error-prone.

Example: [compute.toys](https://compute.toys) has a prelude injected in all shaders. It would be nicer if shaders had a `import computetoys::*;` by default instead.

---

personally I would love this functionality built-in:

<details>
  <summary>Spoiler</summary>

```rs
// aliases for square matrices.
alias mat2f = mat2x2f;
alias mat3f = mat3x3f;
alias mat4f = mat4x4f;
alias mat2<T> = mat2x2<T>; // this is not possible currently
alias mat3<T> = mat3x3<T>;
alias mat4<T> = mat4x4<T>;

// missing built-in functions present in glsl
fn inverse<T>(m: mat2<T>) -> mat2<T> { ... } // this is not possible currently
fn inverse<T>(m: mat3<T>) -> mat3<T> { ... }
fn inverse<T>(m: mat4<T>) -> mat4<T> { ... }
...

// debugging tools, they are no-op in a "release" environment.
fn runtime_assert(val: bool) { }
fn debugger() { }
...
```

</details>

this is compute.toys' prelude:

<details>
  <summary>Spoiler</summary>

```rs
alias int = i32;
alias uint = u32;
alias float = f32;
alias int2 = vec2<i32>;
alias int3 = vec3<i32>;
alias int4 = vec4<i32>;
alias uint2 = vec2<u32>;
alias uint3 = vec3<u32>;
[...]

Issue #103, opened 2025-03-02
Make sub-modules private by default


stefnotch:
I think we should make sub-modules private[^1]-by-default to help with pre-bundling in wesl-js, to help language servers analyze WESL and encourage library authors write good libraries by default. We want it to be easy to find all (visible) modules.

If I do `import("bar.wesl")`, our wesl-js bundler plugin grabs all wesl files and sends them to the browser. To do that, it needs to know what all modules are.

If I do the following, then the language server needs know what all *visible* modules are.

![Image](https://github.com/user-attachments/assets/949d7753-b6c2-406f-88ea-f5cccb2f92ba)

In a library, we just assume that all `.wesl` files are modules.

In the user project, this involves scanning the filesystem. This is where we currently have globs.

However, if we make sub-modules private by default, then it becomes easy to discover all modules. 

1. Start with the root (public)
2. Follow the imports[^2] (definitely required)
3. Done

To expose a sub-module, a person would have to write
```rs
@public
import self::lighting;

// ^ I made up some syntax
```

We originally rejected this private-by-default idea for two reasons
- It adds some extra typing. Every module needs to explicitly expose its public children.
  - I believe that this is *desirable* for library authoring. Good libraries want to carefully control their public API.
  - But it is still extra work in user code.
- Private-by-default for everything makes WGSL files unusable. This proposal doesn't suffer from that issue.

[^1]: More specifically, it should be `pub(super)`. As in, only visible to the parent module.
[^2]: I'm assuming that this also includes the public submodules.

stefnotch:
There are two places where a module with only private items is different from a private module.

When I do a re-export, I can re-export such a module
```
public import package::foo::bar::internal_module;
```

And if alias-everything becomes a thing, then the same kind of code can be written
```
alias nya = package::foo::bar::internal_module;
```

From a language server perspective, that means that *at least in those cases*, I have to suggest the `internal_module` in the autocomplete suggestions.

Issue #98, opened 2025-02-23
allow marking elements as private within a module


mighdoll:
Shader projects with lots of modules and/or lots or people will want to define some functions as private, for all the typical information hiding / modularity reasons. 

Let's add a way to mark some elements as private (inaccessible from `import` statements). (We've decided in #28 to make elements public be default.)

I suggest we use the `private` keyword, but we could also consider an annotation like `@private`..

stefnotch:
The annotation would have two interesting properties
- Parsing it is already done. No changes needed.
- It'd be valid only above some global declarations. Everywhere else it'd be invalid WESL.

mighdoll:
We might want to check with the wgsl folks when we get to this. I guess `private` would be more likely what wgsl would eventually adopt but I'm not sure.

stefnotch:
Relevant:
https://github.com/wgsl-tooling-wg/wesl-rs/issues/9#issuecomment-3143904998

stefnotch:
Reposting two of my messages from Discord: Private by default is an obvious choice

> Regarding visibility (public/private), I keep trying to find a balance between how people typically write shader code today, and the big vision of "shader code should work effortlessly with host code". Which projects like Rust-GPU and Cuda are also trying to achieve.
> 
> And now it occurs to me that GLSL with "public by default" is actually the odd one out. 
> - Slang: `internal` (aka "visible within the package") <https://shader-slang.org/slang/user-guide/modules.html#access-control>
> - Metal and HLSL: Pretty sure that they roughly follow C++'s "classes are private by default" 
> - Cuda: Ditto

But WESL should remain WGSL compatible. Wouldn't it be annoying if WESL locked you out of using WGSL code?
Well

> Let's put ourselves in the shoes of WGSL language designers. `.wesl` files don't exist. 
> If we add `import` statements, then we have a choice
> - You can import anything from other shader files
> - You need to edit existing shader files to have `export` statements 
> 
> Neither of those choices is a breaking change. Both just allow for more features. 
> 
> And then we take a look at how import statements were introduced to existing languages. Examples are ES Modules in JavaScript and C++ modules. 
> And we notice that both of them totally added an `export` keyword. And apparently the community accepted that cost.

Which is a lot of words to say "adding visibility to an existing language is fine, just based on prior precedent".

Link to more discussions https://discord.com/channels/1275293995152703488/1309954314961289256/1437203585640435723

stefnotch:
[...]

Issue #65, opened 2025-01-14
'@publish import' and explicit control of the host visible shader interface


mighdoll:
From [discord discussion](https://discord.com/channels/1275293995152703488/1297099821303201803/1313598305191727135):

WGSL statements that change the host visible interface of the shader, like `override` or `@compute`, should only be linked if they're referenced from the root module (e.g. `./main.wesl`).

Host visible names are: override, entry points (@compute, @vertex, @frag), and resource binding numbers (@group(0) @binding(1))
- host visible names are only linked as host visible names if they are:
	- declared in the root module, or
	- if they are explicitly mentioned in the root module with in an `@publish import` statement like:
		- `@publish import super::util::over;` 

Error conditions:
- it's a link error to use `@publish import` in any module other than the root module.
- `override` elements declared in non-root modules and not imported via with `@publish import` are converted to const elements. (or simply dropped if they are not referenced even indirectly from a root module element)
- it's a link error if a non-root module resource binding is referenced indirectly from a root element but not made visible with `@publish import` in the root module. (but it's ok to have non-root module resource binding that's unreferenced.)

stefnotch:
We want explicit `@publish`, so that our import statements are side-effect free. Just doing an import does nothing.
One has to *use* it for it do something.

 (Just like we're doing with const_assert)

stefnotch:
Should we cut this from M1? I think it's not much of an ask to tell people to just write their host visible shader interface in the root module.

mighdoll:
+1 for the slimming-M1 impulse.

Perhaps communicate in the spec what we want people to do for M1 and link to this as a future issue? 

Or if we're actually in agreement on this design and just awaiting implementation, then spec it this way and asterisk the implementations for M1+ / more likely M2.

stefnotch:
I'm moving this to M2, since I'd like this `@publish` to have the same shape as our visibility qualifiers. (if possible, that is)

For M1, we don't have to do anything at all. Neither our wesl-js nor wesl-rs seem to enforce anything about where entry points and pipeline overridable constants are allowed.

I added a bit to the wiki about recommending that they be in the root module
https://github.com/wgsl-tooling-wg/wesl-spec/wiki/Authoring-Shaders#shader-entrypoints-and-co

ncthbrt:
Have you considered aliasing as a means of publishing?

stefnotch:
That is a really good point, we haven't considered that.

[...]

PR #187, opened 2026-05-12
visibility design and discussion

mighdoll:
TLDR

1. **Module visibility.** Three levels: `public` (visible to any package),
   `private` (declaring module only), and *package* (same package, the unmarked
   default).
2. **Re-exports.** `public import` re-exports the imported names under the
   current module's path; `public import path::*` does the same for a wildcard
   (re-uses `@wildcardable` diagnostics).
3. **Host pipeline visibility.** The link's root module defines the host-visible
   API: entry points, resource variables, and pipeline-overridable constants
   reach the host only if the root module declares them or `public import`s them.
4. **`package.wesl`.** The file backing a package's top-level module, so its
   items are reachable as `<package>::item` (e.g., `import wgsl_test::*`). One
   place a library can put a re-export prelude.

Why these four things are covered:
- module visibility is needed for re-exports
- re-exports are needed for wildcards
- pipeline visibility works best integrated with the module visibility design
- package.wesl is so that we can write e.g., `import wgsl_test::*;`

(Note that VisibilityLanguages.md is discussion background, it'll be removed before merging.)

### Would resolve:

* [#65 @publish import](https://github.com/wgsl-tooling-wg/wesl-spec/issues/65)
* [#98 allow private marking](https://github.com/wgsl-tooling-wg/wesl-spec/issues/98)
* [#28 public by default](https://github.com/wgsl-tooling-wg/wesl-spec/issues/28) (closed)
* [#130 wildcards](https://github.com/wgsl-tooling-wg/wesl-spec/issues/130)
  (largely covered already; this PR fills in the re-export piece)
* [PR #62 comment for lib.wesl](https://github.com/wgsl-tooling-wg/wesl-spec/pull/62#discussion_r1945348373)

### Related

* [#103 sub-modules private by default](https://github.com/wgsl-tooling-wg/wesl-spec/issues/103)


s-ol:
This looks very solid to me and would definitely bring a big usability gain to libraries like lygia. The one thing that doesn't seem be discussed in much detail is `package.wesl` - from what I gather the package structure becomes something like:

- pkgname/
  - package.wesl
  - mod_a.wesl
  - mod_b.wesl
  - mod_b/
    - nested.wesl
  
It seems like the docs here don't explain why this approach is chosen over an `index`-style approach, i.e.:

- pkgname/
  - index.wesl
  - mod_a.wesl
  - mod_b/
    - index.wesl
    - nested.wesl

Personally I don't have a strong opinion either way, but there is lots of prior art for the second option (most JS package systems, Python, Rust, Lua) while I can't think of anything that has this strict root/non-root separation off the top of my head.

stefnotch:
I believe that we should be able to dig up the Rust recommendations for that. IIRC, Rust discourages the `mod.rs` style, because it leads to a lot of files with the same name. Which means that whenever a tool only or primarily shows a file name, one ends up with very little useful info.

For example, if I have 3 tabs open in VSCode and they all say `mod.rs`, I will have a hard time finding the right one.

stefnotch:
Also quick and very inconvenient note: `private` is somehow not a reserved WGSL word.

stefnotch:
`public import path::*` has one notable technical difficulty: It makes import resolution a lot more complex.

Without wildcards, resolving an import means going to the right module and finding the item.

With wildcards, resolving an import means looking through a *set of* modules to find the right item.

With wildcard re-exports, resolving an import means looking through a set of modules *recursively* to find the right item. And since modules can have cycles, one now needs to keep track of which modules have been visited during import resolution. In the language server, this is a fair amount of extra complexity.

This particular feature also, I believe, interacts really badly with macros or other compile time code generation. Rust's import system ends up being a very complex fixed point finding algorithm which may sometimes simply fail.

(I think it also interacts with parameterized modules. Since wildcard reexports is the missing piece to make import statements recursive. If imports can cause any sort of computations, which compile time generated code does, and which parameterized modules do, then one gets the halting problem. Or so the story goes.)

Usually I wouldn't object to complexity too much.
In this case, I did want to point it out, since wildcard re-exports enables nothing new. It is a very rarely used feature and I could always manually write it out.

mighdoll:
> Personally I don't have a strong opinion either way, but there is lots of prior art for the second option (most JS package systems, Python, Rust, Lua) 

> while I can't think of anything that has this strict root/non-root separation off the top of my head.

Not sure, always interesting to look around tho! 

Here I was only filling the gap where our current scheme doesn't allow something at the root level. I think small libraries especially might like to enable their users to `my_pkg::foo();` or `import wgsl_test::*;` without adding an extra path level.

It'd be a separate issue, but your comments about consistency... I suppose we could revise the existing rules so that `X::X.wesl` always flattens the contents to `X::`. `package::package.wesl` flattens to `package::`. `lygia::math::math.wesl` flattens to `lygia::math`, etc.

mighdoll:
> Also quick and very inconvenient note: `private` is somehow not a reserved WGSL word.

wat!? good catch. Perhaps `@public` `@private`. I think we'll want to ask the WGSL committee too.

mighdoll:
> Usually I wouldn't object to complexity too much.
> In this case, I did want to point it out, since wildcard re-exports enables nothing new. It is a very rarely used feature and I could always manually write it out.

Worth thinking about. I like the idea of implementing the needed features w/o gilding the lily too much. 

The cycles/complexity concerns seem heavy for a feature that isn't must-have. And foreclosing options for parameterized modules would be unfortunate.

Also I realize that the PR doesn't address some related issues:
- what if there's a latent collision between wildcard re-exports in a library? That would be something we'd want to catch at publish time, not wait for clients to use the library and find an error.
- what happens if you wildcard re-export from a module with wildcard imports?

I'm thinking best would be to pull wildcard re-exports from this PR. 

Side note, if we go with the publish tool route as planned with #183, I suppose we could also expand wildcard re-exports like we're planning for wildcard imports. That would help for the cross package cases.

mighdoll:
pushed a draft
- `@public` and `@private`
- wildcard re-exports deferred

stefnotch:
We could also go with
- `pub`, `priv`
- `public`, `protected` (Rust does not have a private keyword, upside of private by default)

Also, what does private mean? Rust goes all in on the hierarchical module system and says
> If an item is private, it may be accessed by the current module and its descendants.
> https://doc.rust-lang.org/reference/visibility-and-privacy.html

mighdoll:
>     * `pub`, `priv`

I like this option too, a bit more cryptic but if perhaps best if we prefer keywords to attributes.

>     * `public`, `protected` 

I like `public` clear and readable. I fear confusion around `protected` - it has a contrary meaning in other languages visibility systems.

I suppose we could do `public` and `priv`..

> Also, what does private mean? Rust goes all in on the hierarchical module system

just the module, no hierarchy. we could expand hierarchically later, it's a nice idea. But I think of most WESL programs as smallish, so perhaps unnecessary complexity, better to defer.

stefnotch:
For the `public`, `protected` option, I was trying to suggest having the following visibility modifiers
- `public` (keyword)
- `protected` (keyword)
- `private` (the default, thus does not need a keyword)

mighdoll:
>     * `private` (the default, thus does not need a keyword)
Ah, that was my first instinct too. I like modularity. But I've come to think that package by default for WESL is better. See `Why package by default?` in the design discussion doc. package by default is low ceremony for most projects, and works better for wgsl compat at the root module..

stefnotch:
Also interesting tidbit:
Rust allows for two different styles of visibility. One is a local and hierarchical style, where I only use public and private. And I apply that to both items and modules. A internal (protected) item is modeled by having a public item inside a private module.

The other style is the global style, where I use 3 access modifiers. Public, internal (protected) and private. 

And apparently some of the very prolific Rust contributors, including epage who maintains a bunch of extremely well written libraries, vastly prefer the global style. So we are on the right track!

https://www.reddit.com/r/rust/comments/1k5yv6p/two_ways_of_interpreting_visibility_in_rust/

stefnotch:
More from Rust land:
The reasons for why they went with private by default for *fields* is to allow for [local reasoning](https://cliffle.com/p/dangerust/2/#local-vs-global-reasoning) and 

> Visibility is often how soundness is achieved for many types in rust. 
> https://rust-lang.github.io/rfcs/0001-private-fields.html

They also include statistics which says that `private` is more common than `public`.

stefnotch:
And from Kotlin lang: They went with public by default, only to realize that it might have been a mistake. So defaulting to a more restrictive choice like "internal" seems wise

https://kotlinlang.org/docs/api-guidelines-simplicity.html

stefnotch:
Note from Java:
Java's default is actually somewhat similar to Rust. In Java, a "package" corresponds to a folder. And classes are visible within a package by default.

stefnotch:
We really should ask the WGSL committee regarding the `private` keyword. Would you like to do the honors?

stefnotch:
I dug into it and asked the WebGPU commitee https://github.com/gpuweb/gpuweb/issues/6264

I couldn't quite figure out *where* it got lost though. It seems to have happened [quite a while ago, judging from Naga's source code](https://github.com/gfx-rs/wgpu/commit/8f1d82f0d2a10052e43935336b5a325ddaa40f9d)

stefnotch:
I talked to a friend regarding this. `priv` would be a weird choice. As much as it pains me to say this, I would prefer `private` as a context sensitive keyword over `priv`.

stefnotch:
`private` is intentionally a context sensitive keyword said David Neto. We can use it, but it'll require a bit of context sensitive parsing.

And Jim Blandy pointed out that Rust doesn't have a private keyword.

stefnotch:
I asked the Slang team! 
[Here's their response](https://discord.com/channels/1303735196696445038/1505624625856643072/1505624656835645745) and the Discord invite link is https://khr.io/slangdiscord. I'm happy to repost the thread here if someone wants me to.

Imports.md
```diff
@@ -85,8 +85,8 @@ import_collection:
 ```
 
 Where `translation_unit` and `ident` are defined in the WGSL grammar.
-`ident`s must not be current WGSL keywords. `ident`s also must not be 
-current WESL keywords: `as`, `import`, `module`, `package`, `super`, or `self`. 
+`ident`s must not be current WGSL keywords. `ident`s also must not be
+current WESL keywords: `as`, `import`, `module`, `package`, `private`, `public`, `self`, or `super`.
```
stefnotch: Remove `private` from this list, it's a context sensitive keyword

Imports.md
```diff
@@ -125,22 +128,18 @@ import a::c::e as f;
...
 
 1. We start with the first segment.
-    * `super` refers to the parent module. Can be repeated to go up multiple parent modules. Exiting the root is an error.
     * `package` refers to the top level module of the current package.
+    * `super` refers to the parent module, so `super::sibling` reaches a module alongside the current one. Can be repeated to climb further; climbing out of the current package is an error.
     * `ident` must be a known package, usually found in the `wesl.toml` file. It refers to the top level module of that package.
 2. We take that as the "current module".
 3. We repeatedly look at the next segment.
-    1. Item in current module: Take that item. We must be at the last segment, otherwise it's an error.
+    1. Item in current module (declared or re-exported via `public import`): Take that item. We must be at the last segment, otherwise it's an error.
```
stefnotch: Can `public import` also re-export modules?

Imports.md
```diff
@@ -125,22 +128,18 @@ import a::c::e as f;
...
-    4. Else go to `current module path/ident.wesl`
+    3. Else go to `current module path/ident.wesl` (or `.wgsl` if `.wesl` is not found)
        * File found: We take that file as the current module.
        * File not found: We assume an empty module as the current module, and continue with that.
-       * (Re-exporting changes the path.)
-       * (Inline modules do not have a path.)
 
-To get an absolute path to a module, one follows the algorithm above. In step 1, one takes the known absolute path of the `super` module, or the package.
-The absolute path of the `super` module is always known, since the first loaded WESL file must always be the root module, and children are only discovered from there.
+The steps above resolve a path; whether the referencing module may then use the result is governed by [visibility](Visibility.md).
```
stefnotch: How does this interact with wildcards?

Is this a name clash?
```
import foo::*; 
import bar::*;

const B = A;
```

when
```
// foo.wesl
private const A = 3;
```

```
// bar.wesl
public const A = 3;
```

Or is visibility more intertwined with name resolution?

Visibility.md
```diff
@@ -1,104 +1,257 @@
...
+the global declaration forms WGSL accepts attributes on (`global_variable_decl`,
+`global_value_decl`, `function_decl`, `struct_decl`, `type_alias_decl`); a
+`const_assert_statement` declares no name and accepts no visibility attribute.
+At most one visibility attribute may appear on a single declaration.
+
+See the
+[WGSL recursive descent grammar](https://www.w3.org/TR/WGSL/#grammar-recursive-descent)
+for WGSL's full attribute production.
+
+### Referring to less-visible declarations
```
stefnotch, line 81: Is there a reason why this should be allowed at all?

Visibility.md
```diff
@@ -1,104 +1,257 @@
...
+
+Visibility governs where a declaration can be named, not which other
+declarations it may mention. An `@public` declaration may name a less-visible
+one: an `@public fn` can return or accept a *package* struct, and an `@public
+struct` can have a field of a less-visible type. A consumer that cannot see the
+type can still use a value of it (call the function, read its fields) but cannot
+name the less-visible type to declare a variable, import it, or construct a new
+value.
+
+WESL publishing tools should warn when an `@public` declaration mentions a
```
stefnotch: What about other WESL tools like the language server?

Visibility.md
```diff
@@ -1,104 +1,257 @@
...
+names to the declaration site, and the declared name (for example
+`my_lib::types::Mesh`). A re-export's reachable name (the last segment of the
+re-export path, or its `as` alias) can differ from the declared name, but
+identity follows the original declaration. Two paths that resolve to the same
+identity collapse with no name collision and no duplicated emission.
+Re-exporting one item under several aliases (`@public import super::types::Mesh
+as MeshA;` alongside `... as MeshB;`) just adds reachable names; all of them
+resolve to the single declaration.
+
+### Only public items can be re-exported
```
stefnotch: Why restrict re-exporting to `@public`? I find it quite useful within a package as well, once the package crosses a certain size threshold.

Visibility.md
```diff
@@ -1,104 +1,257 @@
...
+
+### Pipeline-overridable constants
+
+A [*pipeline-overridable constant*](https://www.w3.org/TR/WGSL/#override-decls)
+is an `override` declaration the host may set at pipeline creation. An override
+is in the pipeline-visible API only if the root module declares it or
+`@public import`s it.
+
+An override with a default value that is absent from the pipeline-visible API
+silently degrades to a constant: the linker bakes in the default and the host
```
stefnotch: This is technically not possible. Consider the following

```
public override bar = 3;

private override foo = bar;
```

Then if we downgrade foo, we get

```
override bar = 3;

const foo = bar; // error, this is an override-expression, not a const-expression
```

https://www.w3.org/TR/WGSL/#override-expressions

VisibilityDesign.md
```diff
@@ -0,0 +1,263 @@
...
+into a WESL project would have its visibility silently determined by how its
+names happen to be cased.
+
+## Why re-exports cannot widen
+
+A `@public import` cannot widen visibility: re-exporting a *package* or
+`@private` item as `@public` is a hard error.
+
+The reason is that the visibility attribute on a declaration is meant to be a
+local contract. An author looking at a *package* item should be able to rely on
```
stefnotch, line 196: I really like this "local reasoning" argument. 👍

VisibilityDesign.md
```diff
@@ -0,0 +1,263 @@
...
+extra `@` on every visibility marker); shaders are already attribute-heavy
+enough that this fits the local style.
+
+WESL does not pin a position for the visibility attribute in the grammar, just
+as WGSL does not pin a position for `@group` or `@workgroup_size`. By convention
+(a formatting rule, not a grammar rule), the visibility attribute sits last,
+immediately before the declaration keyword, so a reader scanning down the left
+margin can find it without hunting through the attribute list.
+
+### Alternatives considered
```
stefnotch: I think we have to go down the route of `public`, `private`[^1]. Making it an attribute is odd for users. 

[^1]: Or `public`, `internal`. Or some other flavour of keywords. But definitely keywords, even if they require context sensitive parsing.

VisibilityDesign.md
```diff
@@ -0,0 +1,263 @@
...
+
+### `@package` attribute on declarations
+
+A `@package` attribute would make the middle level explicit on a declaration,
+rather than leaving it implied by the absence of `@public` or `@private`. The
+unannotated form covers every present use, so the attribute isn't part of the
+current design; it would mainly help where an author or a tool wants to see that
+a `package` level was chosen deliberately rather than left unmarked.
+
+### Submodule re-export
```
stefnotch: I think Bevy will want those. But I'm happy to leave it for a future PR.

VisibilityDesign.md
```diff
@@ -0,0 +1,263 @@
...
+
+### re-export widening from root
+
+A plain `.wgsl` file with an entry point cannot be aggregated into a root module
+via `@public import` without modification (see
+[Aggregating entry points](Visibility.md#aggregating-entry-points)). A
+relaxation would let the root module `@public import` *package* items from the
+same package, since the root defines the package's external and pipeline-visible
+API. The cost is loss of orthogonality: what `@public import` accepts would
+depend on whether the importer is the root.
```
stefnotch: Ah, so basically I cannot do `public import package::some_plain_wgsl_file::fragment_shader;`. Because the `fragment_shader` in the plain WGSL file isn't public.

stefnotch: But I should be able to do `internal import package::some_plain_wgsl_file::fragment_shader;`. After all, package-level items are pipeline-visible.

VisibilityDesign.md
mighdoll: agreed, will change back. Seems pretty parseable.

VisibilityDesign.md
```diff
@@ -0,0 +1,263 @@
...
+  helper modules. The author already knows the whole codebase, so the
+  disciplined boundaries that pay off in larger code don't add as much value.
+
+* **Don't force visibility decisions on every author.** Any non-trivial module
+  shares at least some declarations with other modules in the package (helpers,
+  types, constants). With private as the default, every author would have to
+  mark those items explicitly, even authors who don't otherwise care about
+  encapsulation. Package as the default makes encapsulation discipline available
+  to projects that want it (via `@private`), without forcing typical small
+  projects to do the extra work.
```
stefnotch: I don't think this argument holds up. Pretty much every host language (C, C++, Rust, Typescript, ...) already makes the authors do this work. And my the dream shading language would be *whatever host language I use*.
So using the same patterns across my codebase is *less* mental overhead for me. Instead of having to remember "wait, well written WESL means that I have to write `private` more often"

Visibility.md
mighdoll: simplicity, could add later I agree

Imports.md
mighdoll: could be a future thing, perhaps..

Imports.md
mighdoll: good question, I think it should be only be a clash if both are visible.

VisibilityDesign.md
```diff
@@ -0,0 +1,263 @@
...
+
+* **WGSL files can be imported without modification.** WGSL has no visibility
```
stefnotch: I sort of agree with this one. It should be possible to incrementally adopt wesl.

But at the same time, I'd also like to note that that existing WGSL code has no imports and is not always designed for it. 

If I were to add visibility to WGSL itself, I would likely choose an option where one *has to* add visibility markers before being able to use the code in other files. This is analoguous to what Javascript did with the `export` keyword.

In fact, as an imaginary WGSL (not WESL) standard writer, I would argue that allowing arbitrary imports is a footgun. Existing WGSL code has been written without a module system in mind. So there are no encapsulation boundaries anywhere. I would start accessing all the right and all the wrong things, without the language steering me in the correct direction.

With my IDE hat on, I would agree. If I add "allow imports from anywhere" into any of my existing WGSL codebases, I'd get a hundred silly recommendations. Often I'd get 5 separate recommendations for the same type, because I copied it 5 times. And worse yet, of those 5 times, 2 of them would be subtly different, due to reasons like https://github.com/gpuweb/gpuweb/issues/4574

VisibilityDesign.md
mighdoll: I like the idea, then bigger projects can have a lint rule that enforces that every declaration has a visibility. But agree also that we should save for a future PR to see what bigger projects want in practice.

VisibilityDesign.md
```diff
@@ -0,0 +1,263 @@
...
+  WGSL file would need to be edited or re-annotated before its declarations were
+  reachable, which would undermine the WESL goal of working smoothly with
+  existing WGSL code.
+
+  WESL could distinguish WESL from WGSL by file extension and give each a
+  different default. WESL doesn't: it weakens the "WESL is a superset of WGSL"
+  principle, and shader source loaded as a string (without a file extension)
+  would still need some other way to mark the dialect.
+
+* **Root module entry points would need explicit markers.** A root module's
```
stefnotch, line 97: This is a solid argument. So "private by default" with a file-level private would immediately break all WGSL files.

Which means that the WGSL standards commitee could not choose this option without introducing a breaking change to WGSL.


In terms of other options:
- For the entrypoints and bindings, there is the option of saying "`@fragment` implies `public`" and "`@group` implies `public`". But that feels weirdly inconsistent with the rest of the language. And this rule fails to cover overrides.
- Root modules with a different behaviour seem weird. Having a consistent language means fewer things that I have to remember.
- `private` could simply not affect pipeline visibility. But then we break encapsulation in a weird way. 
- `private` could be much less strict, similar to Rust. Rust made the observation that file level private is often still too strict.

VisibilityDesign.md
```diff
@@ -0,0 +1,263 @@
...
+### Why not private by default?
+
+Making private the default favors strict encapsulation: every cross-file use
+requires an explicit visibility marker. That discipline is attractive for
+libraries with stable APIs and large codebases. It's less of a fit for typical
+WESL projects, which tend to be small and internal. Four arguments push WESL
+toward package as the default:
+
+* **Encapsulation gains are smaller at WESL scale.** A typical WESL project is a
+  handful of files: a vertex/fragment pair, a compute shader, perhaps a few
```
stefnotch: I'm not sure if I agree with the assumption here. I'm more familiar with game development, where a typical shader codebase is
- a few files that I wrote. These are indeed somewhat small. But they are also very *standalone*, due to the nature of things. In fact, while many big game engines (e.g. Unreal Engine) have "material functions", it is kind of rare for me to import code from another shader that I wrote.
- a massive amount of game engine shader code that I rely on. I import a few types from here and the rest is a black box

And I suspect that the same is true on the web, since frameworks like Babylon.js and Three.js are quite popular.

So targeting small shader projects that are not already a part of a game engine is, I think, targeting an overly small part of the userbase.

Visibility.md
stefnotch: I think it'd be simpler to implement if we have the internal variant as well. The way it's currently phrased with `@public` means that I need to add quite specific checks for the restrictions.

(This is very speculative of course. My point is mostly that I think we can safely allow more here, without adding notable implementation complexity.)

VisibilityDesign.md
stefnotch: Good, then we agree! Bevy can have their cake and eat it too, but first we gotta bake the cake. 

So this should be a follow up PR once we're sure that we need it.

Visibility.md
mighdoll: Authors can write a fn that makes and returns a struct with private visibility. API users can pass the struct back to another fn, but can't build one them themselves w/o going through the API. That might be nice for the author, maybe they want to track every struct built in a table or something and don't want users to be able to able to cheat around their intended `makeThing()` api.

Requiring that every struct mentioned in a public fn also be public is reasonable, but less flexible for API authors. Other languages undoubtedly differ on this. I was just scanning through Rust [RFC 2145](https://rust-lang.github.io/rfcs/2145-type-privacy.html) which makes it sound like Rust had a strict requirement and then changed to warn/lint. Go allows it but has a lint. TypeScript allows it I think.

Visibility.md
mighdoll: good point, other tools like the lang server should warn too.

Visibility.md
stefnotch: Couldn't one achieve the same result by making one of the struct fields private? Then the struct is also not constructible for the API users.

I don't fully understand why Rust relaxed their rules, from the sounds of it they started out with a suboptimal version of "no private in public".

Visibility.md
mighdoll: Good one. It won't work to simply convert to const. 

We could require that all statically reachable overrides not be private. But private variables derived from overrides seem like a good feature for modularity reasons. (And users can't have a private const derived from an override as you point out.)

As an alternate implementation, the linker could mangle the override variable name but leave it as an override (for statically used private overrides) without rewriting to const. A linker could optimize and convert to const for simple cases, perhaps add a `__private_` prefix to the mangled override name to warn the user.

I'll change the spec to not prescribe the mechanism, but leave the semantically useful modularity feature of private overrides initialized from public ones.

VisibilityDesign.md
mighdoll: Agreed that many people will use big frameworks when they use webgpu. And the authors of big projects still have all the capabilities they need, regardless of the defaults discussed here. Here it's more about syntax convenience for the _users_ of those frameworks (and those not using frameworks). The point is that most people using WebGPU will be writing smallish things with smallish teams that aren't so important to carefully internally modularize.

Visibility.md
mighdoll: It's a good question, I went back and forth on this as I was drafting. What's simpler for the user? I agree it's a little odd to have to declare an app's own fragment shader as public just so that it can be pulled into the root module. On the other hand it's another kind of re-exporting for users to learn about, which adds complexity and brings up questions about internal re-exporting in general (two paths to a declaration, module path vs declaration visibility, etc.) that I'd rather keep out of scope for this PR.

I'd prefer to keep re-export to public import only for this if possible. We can revisit and extend to package import if need be in a subsequent PR.

VisibilityDesign.md
mighdoll: agree that would be nicely orthogonal, but prefer to [defer](https://github.com/webgpu-tools/wesl-spec/pull/187#discussion_r3300138285) to keep this PR constrained.

VisibilityDesign.md
mighdoll: It's a good challenge, this argument may not land for many. Presuming we end up going with package by default, I'll reorder this section to lead with the structural arguments and de-emphasize this point.

The host-language story is mixed (e.g. Java, Scala, Kotlin, Swift are not private by default, so it's not a clean vote), but for users coming from TS or Rust you're right that this argument doesn't carry much weight on its own.

Visibility.md
mighdoll: > Couldn't one achieve the same result by making one of the struct fields private? Then the struct is also not constructible for the API users.

Struct field visibility is interesting, but perhaps mostly orthogonal.

I agree we could block consumer construction of structs w/ private fields. `Foo(1,2,3)` wouldn't work outside the module if one of `Foo`'s fields was private. But that workaround only matters now if we go strict on the leaked-type rule, and going strict cascades through every transitive field reference. Sounds like Rust wanted to relax that in RFC 2145, so not sure we'd go that way.

i.e. if we eventually add private struct fields, it's plausible, maybe even likely that we'd still want to allow public structs to have fields of private type. In that case, private fields would give a new way to have private struct types: named-but-not-constructable in addition to the current unnamed-not-constructable.

Either way, I'd suggest we defer til we decide to do field level visibility.

VisibilityDesign.md
mighdoll: I suppose the IDE suggestions perspective might cut both ways. If I copy in some useful WGSL code from an old project or the web, then I'd like to use the IDE auto complete to explore the WGSL. But if I leave the WGSL code in for a while, then I agree I'd prefer to just have the one routine I know I need suggested but can't quite remember how to spell and not the other ought-to-be-private declarations. Autocomplete is probably not the deciding factor regardless; the incremental adoption is more important, as you note.

VisibilityDesign.md
mighdoll: > `private` could be much less strict

We could use package by default, and have private cover both the module and its descendant modules.  So a `private fn foo` declaration in module `shaders/bar.wesl` would be visible also to `shaders/bar/zap.wesl`.

Seems reasonable and doesn't conflict with the current default or the import path model. The locality cost for the reader is mild, they'd have to consider their relative place in the module tree, where today they just have to consider the declaration itself. It's additive, we can add it later without breaking anything. Prefer to defer until we know we want this approach vs. other potential mechanisms for bigger projects.

mighdoll: > other options: 

(suggested on discord) Package visibility by default for items, but a submodule is path-private: not reachable to other modules unless its parent forwards it. The forwarding mechanism is re-exports syntax applied to module paths, rather than just items as in the current PR. So a `tests/` directory whose parent doesn't forward it is unreachable from non-test code in the same package.

But: 
- Today an item's visibility is clear from its declaration. Path-private modules mean you'd have to read ancestor files to know what the parent forwarded. Visibility becomes non-local
- A main advantage of package by default is that small projects with a few files don't need to care about visibility rules. This would be effectively private by default. We'd lose the advantages of package by default.

I agree we'll want a solution for `tests/`, but #160 sketches an alternate approach based on wesl.toml, so we've at least one alternate way forward.

Path visibility might be worth a separate PR if #160 doesn't work out. Either way I think we'd still want declaration level private for the routine within-module case. Path visibility wouldn't replace it.

Visibility.md
stefnotch: Sounds good, name mangling seems like the best we can do in the general case.

VisibilityDesign.md
stefnotch: > Today an item's visibility is clear from its declaration. Path-private modules mean you'd have to read ancestor files to know what the parent forwarded. Visibility becomes non-local

There are *semver relevant* cases where making a module private is important. Main one being module re-exports.
If I write a little library that does the following
```
public bevy_pbr::internal_module as bar;
```
Then Bevy's `internal_module` is suddenly part of the public API of my library users. Aka renaming it becomes a breaking change for Bevy, despite the intent very much being "that module should not be public".

Therefore, we do need a way of saying "this module is private".
Which yes, makes visibility less local.

We could also get there with an explicit rule of "a module's visibility is the maximum visibility of all its items". As in, if I only have internal items, the entire module is internal.

stefnotch: > This would be effectively private by default

It is significantly less strong than private by default.
With a strong interpretation of private by default, I have to add a marker to each item.
```
// my_module.wesl
public const a = 3;
public const b = 5;
```

With "sub modules are private by default", I only have to add a marker to the parent module's import statement.
```
// package.wesl
public import package::my_module;
```

stefnotch: > So "private by default" with a file-level private would immediately break all WGSL files
> -- @stefnotch 

Well, looking at Javascript's history, this isn't entirely correct either. JS introduced modules and required the callers to say "this is a module"

That said, WebGPU could do an equivalent thing.
```
const vertexShader = device.createShaderModule({ 
  module: fancyShaderString, 
  // instead of the current 
  // code: shaderString
});
```

Imports.md
stefnotch: https://github.com/webgpu-tools/wesl-spec/issues/189

We can do it as a future thing, but we might need this for Bevy

Visibility.md
```diff
@@ -209,49 +212,51 @@ The re-import does not create a new resource variable; the original
...
+`override` is in the pipeline-visible API only if the root module declares it or
+`public import`s it.
+
+An `override` that is absent from the pipeline-visible API is not host-settable
+and takes its default value. A WESL linker may transpile an `override` with a
+default value to an `override` in WGSL output with a mangled name and no `@id`,
+transform it into a `const` if its default is a const-expression, or inline its
+initializer at use sites. These strategies are observably equivalent to the
+host. If the `override` doesn't have a default value, it is a link error if the
+`override` is statically referenced from the root module.
```
stefnotch: I suppose one way of properly dealing with `override` would be to extend WESL to also cover a small bit of the WebGPU API. We'd extend the way `createShaderModule` works or something along those lines.

VisibilityDesign.md
```diff
@@ -0,0 +1,263 @@
...
+[Why three levels and not two](#why-three-levels-and-not-two)). It would also
+impose a retroactive visibility semantics on existing WGSL identifiers, which
+follow no convention tying capitalization to visibility: a `.wgsl` file dropped
+into a WESL project would have its visibility silently determined by how its
+names happen to be cased.
+
+## Why re-exports cannot widen
+
+A `@public import` cannot widen visibility: re-exporting a *package* or
+`@private` item as `@public` is a hard error.
```
stefnotch: There is one amusing way of bypassing this restriction.

```wesl
//- utils.wesl
struct InternalStuff { a: 32 }

//- package.wesl
import package::utils::InternalStuff;
public alias NowItIsPublic = InternalStuff;
```

mighdoll: clever and amusing, indeed!

Imports.md
```diff
@@ -174,10 +176,22 @@ The [`wesl.toml`](WeslToml.md) file provides linker configuration options affect
...
+the current module) through any subdirectories to the module's `.wesl` or
+`.wgsl` file. The starting directory is always known: each module's path is
+fixed relative to its package's root directory, which the linker takes from a
+`wesl.toml` file or its own configuration.
+
+### `package.wesl`
+
+`package.wesl` is the file backing a package's top-level module, placed at the
+package's root directory. Items declared in or `public import`ed from
+`package.wesl` are reachable as `<package>::item`.
```
stefnotch: I'm currently implementing this in the language server and I would like a small bit of clarification.

This is only allowed for the top level module, right?
Examples would be `root = "./package.wesl"` or `root = "./shaders/whatever/package.wesl"`.

For other modules, it'd be unusable, since `package` is a keyword.

And when there is a `package.wesl` file, it assumes that the children are in the same directory.
So, for example
```toml
root = "./shaders/package.wesl"
```

```
wesl.toml
shaders.wesl <== effectively ignored
/shaders
  package.wesl <== root filed
  bar.wesl <== referenced as package::bar
  bar/
    foo.wesl <== referenced as package::bar::foo
```

Is that correct?

mighdoll: > Is that correct?

Mostly agree! But I'm thinking `wesl.toml` `root` should name the folder, not a `.wesl` file.

> shaders.wesl <== effectively ignored
> /shaders

We never link `.wesl` files outside the root folder. So any peer to `shaders/` is ignored regardless of name.

> This is only allowed for the top level module, right? 

`package.wesl` is special only in the root folder. It's just fine to omit `package.wesl`. In that case the package has no top-level items (nothing to reach as `package::someItem`), but submodules like `package::someModule` still resolve.

> For other modules, it'd be unusable, since `package` is a keyword.

Good point about the `package` keyword constraint. A `package` segment elsewhere wouldn't be legal in a module path (and would invite confusion if it were). The language server could helpfully warn about misplaced `package.wesl` files.

---

> Examples would be `root = "./package.wesl"` or `root = "./shaders/whatever/package.wesl"`.

Thinking about it, I think we can drop the `wesl.toml` spec'd ability to specify a root file (vs. a root folder). A standard name is more discoverable and one mechanism for `root` is simpler. 

> `root = "foo/bar"` is hard to deal with.

Dropping root file in favor of root folder also fixes the hassle you [noted](https://github.com/webgpu-tools/wesl-spec/issues/178#issuecomment-4163239244) on #178, where the ambiguity of whether `bar` is a file or directory creates annoying edge cases.

BTW, I think nesting via sub-directory `wesl.toml` files as in PR #182 / wesl-rs #193 is fine w/all of this. A nested `wesl.toml` is a separate package, and the `root` and `package.wesl` rules above apply per-package. So `package::` in a nested project's file refers to that nested project's root.

stefnotch: Yeah, specifying a root folder also removes the ambiguity and makes it simpler to implement. I can work with that.

stefnotch:
Should we see if we can get #183 over the finish line?

**Edit from the future**
#193 and the new `@!wildcardable` syntax do the trick for me. 


**Old proposal**
I do have one alternative proposal for this and #183, which is based around making full use of import statements. Here I'm only listing the upsides.

It would look like this
```ts
@wildcardable
public import bevy::prelude;
```
Upside: No new `module` keyword for this use-case. Fewer keywords is better.

Sub-modules would be private by default. 
```ts
public import package::lights;
public import package::shadows;
internal import package::utils;
```
Upside: Automodules would no longer be a feature. It adds a lot of complexity to the language server. (I also suspect that the current version of automodules is not implementable in browsers, since it relies on web servers returning a "not found" within a reasonable timeframe.)
Upside: Automodules were an anti-pattern for libraries, since libraries should document their modules. Which cannot be done with automodules.

`@if` would compose nicely. For example, to do a switch between desktop and mobile, I would do
```ts
@if(DESKTOP)
internal import package::desktop::raytraced_lights as lights;
@if(!DESKTOP)
internal import package::mobile::lights;
```
and now `package::lights` always refers to the correct thing. More importantly, the language server will *not* suggest `package::mobile::lights` as the autoimport whenever I use `lights`. I think this is a big upside. The current proposals do not have a good way of doing this.


`internal` becomes a keyword. Thus enabling the following, which does not widen the import
```ts
internal import package::vertex::vertex_main;
```

Upside: Enables in-browser unit test runners.
Currently, if I want to unit test my WESL code, I need to set up wesl-js with a *bundler* for it to discover all `.wesl` files. Otherwise, given a `/shaders/package.wesl` URL, the browser has no way of discovering the `/shaders/tests/foo.wesl` file.
However, if sub-modules are private by default, then it becomes logical to add import statements for them. And thus, the browser can discover the files. This also matches the behaviour of Javascript.


Finally, I took a stab at implementing this. It's about equal in terms of complexity, since this just requires re-exports.

Imports.md
```diff
@@ -125,22 +131,18 @@ import a::c::e as f;
...
 
 1. We start with the first segment.
-    * `super` refers to the parent module. Can be repeated to go up multiple parent modules. Exiting the root is an error.
     * `package` refers to the top level module of the current package.
+    * `super` refers to the parent module, so `super::sibling` reaches a module alongside the current one. Can be repeated to climb further; climbing out of the current package is an error.
     * `ident` must be a known package, usually found in the `wesl.toml` file. It refers to the top level module of that package.
 2. We take that as the "current module".
 3. We repeatedly look at the next segment.
-    1. Item in current module: Take that item. We must be at the last segment, otherwise it's an error.
+    1. Item in current module (declared or re-exported via `public import`): Take that item. We must be at the last segment, otherwise it's an error.
```
stefnotch: New question: If it is an item in the current module, but the item is not accessible, what should happen?

1. Throw an error. This always behaves consistently.
2. Ignore it and continue on. Downside is that name resolution can now depend on where you're importing something from in a very funky way.

stefnotch:
New question: The default visibility of an import statement is private, right?

So it's only visible to the current file.
For example, only `Nya` is accessible from the outside. To access anything else, one has to be in the current file.
```wgsl
// my_file.wesl
import bevy::Bar;
private const Foo = 3;
public const Nya = 5;
```

But what about self-imports?
For example, should this be legal?
```wgsl
// my_file.wesl
import package::my_file::Foo as Bar;
private const Foo = 3;
```

Rust argues that yes, it is legal, because 
```rust
use std::f32::NAN; // creates an item NAN with private visibility

use self::NAN as Nya; // imports NAN, which is allowed, because this is the same file as NAN 
```

mighdoll:
> Upside: Automodules would no longer be a feature. It adds a lot of complexity
> to the language server. 

(We've started discussing in #193; I agree direct resolution is better.)

> ```
> @wildcardable
> public import bevy::prelude;
> ```
> Upside: No new `module` keyword for this use-case. Fewer keywords is better.

Users consume wildcards per-module, so I think it's better to mark wildcardable per-module.  Otherwise we risk inconsistencies between what can be imported directly vs what can be wildcard imported from the same module. That seems likely confusing for users.

Saving a keyword is nice of course, and we found options for that in [#183](https://github.com/webgpu-tools/wesl-spec/pull/183#issuecomment-4393795286). (I just don't think we like them better.)

Let's discuss further in #183 if we should revisit `module` syntax.

> Sub-modules would be private by default.
> ```
> public import package::lights;
> internal import package::utils;
> ```

A module whose items are all non-public is effectively private already, which falls out of item visibility rather than needing a module-private concept.

I'd rather leave future module visibility questions undefined for now. We might not need the separate module-visibility axis.

> `internal` becomes a keyword. Thus enabling the following, which does not widen
> the import
> ```
> internal import package::vertex::vertex_main;
> ```

Yep, we could add a keyword if we want package-internal re-exports.  As to the keyword, the draft suggests `package` rather than `internal` to stay in sync with the `package::` syntax. Do we need package-internal re-exports?  I think it's certainly reasonable, but debatable. (It's in the [Future possibilities](https://github.com/webgpu-tools/wesl-spec/blob/visibility/VisibilityDesign.md#package-import) section now.)

> ```
> @if(DESKTOP)
> internal import package::desktop::raytraced_lights as lights;
> @if(!DESKTOP)
> internal import package::mobile::lights;
> ```

In the current proposal you can conditionally re-export _items_ for the _public_ API:

```
/// lights.wesl
@if(DESKTOP)
public import package::desktop::raytraced_lights::{FooLight, BarLight};
@else
public import package::mobile::lights::{FooLight, BarLight};
```

(I see the ConditionalTranslation spec lags the test suite and wesl-rs/wesl-js on allowing conditions on import statements; we should fix that.)

Also just noting that this kind of thing will be awkward until we have some interface contracts to help keep these desktop vs. mobile APIs cross-compatible.

Re-exports become more ergonomic for library authors if we enable wildcard re-exports too: 

```
/// lights.wesl
@if(DESKTOP)
public import package::desktop::raytraced_lights::*;
@else
public import package::mobile::lights::*;
```

That's listed in the future section in this draft. It helps library prelude _authors_ with maintenance, but doesn't add anything new for _users_ of wildcards.  We can add it later, by having the library packaging tool rewrite wildcard re-exports to named re-exports. The design doc lists a couple of other open questions we'll have to work through (module re-export cycles, etc.; all seem solvable).  Library-safe wildcard re-exports are separable, so better in a separate PR when we want it.

Re-export for package internal APIs would make sense in the current proposal if/when we add a `package import` form (the design doc lists it). I don't think it requires a separate module visibility axis, though it might also motivate the explicit `package` declaration keyword at some point.

> given a `/shaders/package.wesl` URL, the browser has no way of discovering the `/shaders/tests/foo.wesl` file.

That's true, but I think the answer is at the tooling layer, not in visibility or imports.  I agree that a test runner needs some way to find its tests.  It finds them by searching for test files (a manifest, or a filename/path convention), not by traversing imports from `package.wesl`. Traversal wouldn't reach them anyway, since tests import their target shaders, not the reverse. So I don't think submodules need to be private by default to make tests discoverable.

> This also matches the behaviour of Javascript

JavaScript test frameworks typically use tooling to search the filesystem for tests by path/filename suffix. wgsl-test does the same thing currently.

Old school browser-only test frameworks would manually list their tests in a test file, AFAIK.

> Should we see if we can get #183 over the finish line?

Yes! I imagine we'll need this one too (#187) to declare success on #183. (The immediate motivation for visibility was because we needed re-exports to merge items from multiple modules for `prelude::`s.)

Imports.md
mighdoll: > New question: If it is an item in the current module, but the item is not accessible, what should happen?
> 
> 1. Throw an error. This always behaves consistently.
> 2. Ignore it and continue on. Downside is that name resolution can now depend
>    on where you're importing something from, in a very funky way.

Option 1 (error). Naming an item that exists but isn't visible to you should be an error. You asked for that specific item, so you should be told you can't have it.

That doesn't block traversal *through* a same-named submodule, though: if `foo` has an inaccessible item `bar` and also a `foo/bar.wesl`, then `foo::bar::baz` still resolves into the submodule. Only the terminal `foo::bar` is the item (and errors if it's inaccessible). Direct resolution (#193) makes this especially clear.

mighdoll:
> New question: The default visibility of an import statement is private, right?
> 
> So it's only visible to the current file. For example, only `Nya` is accessible from the outside. To access anything else, one has to be in the current file.
> 
> ```wgsl
> // my_file.wesl
> import bevy::Bar;
> private const Foo = 3;
> public const Nya = 5;
> ```

Yes, import statements don't have visibility, only items.  So only `Nya` is importable externally.

> But what about self-imports? For example, should this be legal?
> 
> ```wgsl
> // my_file.wesl
> import package::my_file::Foo as Bar;
> private const Foo = 3;
> ```

Yes, legal: `import package::my_file::Foo as Bar;` is just a local alias, and an item is always visible to its own module.

Note that `public import package::my_file::Foo as Bar;` would be an error. A re-export can't widen visibility.

stefnotch:
I am happy with this direction. A proper review will come later.
I do still think private-by-default might be a better trade-off, but there's nothing strong backing it up at the moment.

So my preferred option would be reviewing and merging this. Afterwards we create a *draft* proposal for private-by-default, to explore what that would imply. For example, a truly good private-by-default proposal would likely involve extending the `createShaderModule` API on the host side.

And the other thing that I'd like to explore is: *What is the most common visibility in any given project.*
I am thinking of the Bevy game engine in particular. That question can be answered once we have a fully Bevy-WESL. In fact, the language server more or less should have all the relevant info, since it will gain the "find all references" capability at some point.


Edit: We talked about this today, the strongest signal will simply be observing what our users do. For example, if people start writing `private` everywhere out of paranoia, then we'll have to tweak the design. Meanwhile if people mostly rely on `internal` being convenient, then the current design is good.
Also, Bevy is a unique project, because its scope is so much larger.

mighdoll:
> I am happy with this direction.
> So my preferred option would be reviewing and merging this. 

We should feel good about the direction, knowing that we can course correct while the language and community are young.

> For example, a truly good private-by-default proposal would likely involve extending the `createShaderModule` API on the host side.

Right that's a hurdle, e.g. for potential upstreaming into WGSL.

>  I am thinking of the Bevy game engine in particular. 

Yeah, having something for Bevy was the original motivation for this. In general visibility control is important for libraries. Also presuming that we don't end up with public by default, we'll either break compatibility or commit to dealing with maintaining migration capability in the tools. So good to get something in soon while there's little to break. More recently, I've realized I would like to use visibility soon for the wgsl-play stuff too (e.g., to make clean root modules via re-exports).

> And the other thing that I'd like to explore is: _What is the most common visibility in any given project._

Just noting we want to look at a breadth of projects / use cases. The design goal I think about most often is how to keep things simple for the common / new user, while enabling power features for the sophisticated.

Visibility.md
```diff
@@ -1,104 +1,255 @@
...
+it declares no name.
+
+### Referring to less-visible declarations
+
+Visibility governs where an item can be referenced or imported, not how
+visible its referrers must be. A `public` declaration may refer to a less-visible
+one: a `public fn` can return or accept a *package* struct, and a `public struct`
+can have a field of a less-visible type. A consumer that cannot see the type can
+still use a value of it (call the function, read its fields) but cannot name the
+type to declare a variable, import it, or construct a new value.
```
k2d222: editorial: would be good to add a couple wgsl examples here.

Visibility.md
```diff
@@ -1,104 +1,255 @@
...
-Exports can also be added as an attribute to the item itself.
+After this, importers can write:
+
+```wesl
+import my_lib::prelude::dot2;
+import my_lib::prelude::M;
+import my_lib::prelude::*;       // wildcard pulls all of the above
+```
+
+A bare `import` brings names into local scope without re-exporting them.
```
k2d222: editorial: with a private visibility. i.e. it behaves like a private local declaration.

Visibility.md
```diff
@@ -1,104 +1,255 @@
...
+
+### A re-export cannot widen visibility
 
-```wgsl
-@export
-struct Foo {
-    x: f32;
+A `public import` re-exports a name at `public` visibility. The original
+declaration must already be `public`; re-exporting a *package* or `private` item
+as `public` would widen its visibility, and is an error.
```
k2d222, line 167: I think it'd be good to allow re-exports of package declarations. This is really useful when you don't want to leak implementation details, e.g. you want your public-facing API to be all in one module, but internally you have implemented things in sub-modules. 

For instance, Lygia has a module for each function. But I don't want to `import lygia::math::mod::mod`. I want `import lygia::math::mod`. `math.wesl` acts as a "barrel file" (JS) and simply re-exports from `mod.wesl`. The declaration has package visibility. I could mark the declaration public, but then from the outside I can access it with two distinct paths, and I don't want my users to even know about the `mod` module! 

In short, publishing a declaration also has the side-effect of publishing the module that contains it.

Visibility.md
```diff
@@ -1,104 +1,255 @@
...
+brings an item into local scope but does not add it to the pipeline-visible API.
+
+A resource variable or `override`
+[statically accessed](https://www.w3.org/TR/WGSL/#statically-accessed) from the
+root module's dependency graph but absent from the pipeline-visible API is a
+link error (except for a defaulted `override`; see below). The check starts from
+declarations in the root module and follows references transitively.
+
+An `import` statement is not itself a static access. An imported item is
+statically accessed when an identifier referring to it appears in a body, type,
```
k2d222: editorial: ...appears in a *statically accessed* body... or attribute of a statically accessed declaration.

Visibility.md
```diff
@@ -1,104 +1,255 @@
...
+initializer, or attribute.
+
+Pipeline-visible items keep their exposed names in the linked WGSL output. A
+root-declared item keeps its declared name. An item re-exported by `public
```
k2d222: editorial: line jump messes up formatting

Visibility.md
```diff
@@ -1,104 +1,255 @@
...
+public fn pbr_fragment() -> @location(0) vec4f { ... }
+
+// app/main.wesl  (root)
+public import pbr_lib::passes::pbr_fragment;    // host can select this entry point
+```
+
+### Resource variables
+
+A [*resource variable*](https://www.w3.org/TR/WGSL/#resource-interface) is a
+`@group/@binding var<...>` declaration that lets host code provide values to the
```
k2d222: editorial: "provide values" -> "bind resources"

Visibility.md
```diff
@@ -1,104 +1,255 @@
...
+shader (uniforms, storage buffers, textures, samplers). Resource variables have
+no host-side fallback, so a resource variable statically accessed but absent
+from the pipeline-visible API is a link error.
```
k2d222: editorial: You already said so in sec. "Pipeline visibility", I think you can merge/remove the other paragraph.

Visibility.md
```diff
@@ -1,104 +1,255 @@
...
+### Pipeline-overridable constants
+
+A [*pipeline-overridable constant*](https://www.w3.org/TR/WGSL/#override-decls)
+is an `override` declaration the host may set at pipeline creation.
+
+A defaulted `override` that is absent from the pipeline-visible API is not
+host-settable and takes its default value. A WESL linker may emit it as an
+`override` with a mangled name and no `@id`, transform it into a `const` if its
+default is a const-expression, or inline its initializer at use sites. These
+strategies are observably equivalent to the host. A non-defaulted `override`
```
k2d222: nitpick: if the override is transformed to a const or inlined, the linker should also check that it was not used in const-only contexts, e.g. in an array size expression. Otherwise, the transform makes invalid code become valid, and it's an observable effect.

Visibility.md
```diff
@@ -1,104 +1,255 @@
...
+// app/main.wesl  (root)
+public import filter_wgsl::blur;
+// error: `filter_wgsl::data` is statically accessed but not pipeline-visible
+// fix: add `public import filter_wgsl::data;` to the root module
+```
+
+The `public import` does not create a new resource variable; the original
+`@group(3) @binding(0)` annotations carry through unchanged.
+
+### Pipeline-overridable constants
```
k2d222, line 257: I think an override should always be public, because it never makes sense otherwise. If it's non-public and initialized, it should just be a const. If it's non-public and non-initialized, it must be unreachable, otherwise it's a link error. 

The same idea goes for bindings and entrypoints. They should at the very least not be private.

... Or is this a warning better suited to a linter/language server? wdyt?

GLOSSARY.md
k2d222, line 1: later, we can add *link error* to this list.

Imports.md
```diff
@@ -137,67 +145,164 @@ import foo::*;
...
-To get an absolute path to a module, one follows the algorithm above. In step 1, one takes the known absolute path of the `super` module, or the package.
-The absolute path of the `super` module is always known, since the first loaded WESL file must always be the root module, and children are only discovered from there.
+Each flattened import statement binds one name in the importing module: the
+last segment of its *import path*, or its `as` alias. The binding is a
+shorthand for the import path; the import path itself need not name a module or
+declaration. Each *reference*, a use of a name in code
+(such as in a function call, type, attribute argument, or other expression),
+determines a declaration path: a bound name expands to its import path, and
+any `::` segments written after it extend the path
+(see [Inline Usage](#inline-usage)).
```
k2d222, line 160: unrelated to this PR, but can import statements refer to other import statements?

```wgsl
import foo::bar;
import bar::baz; // foo::bar::baz or bar::baz?

File: Visibility.md
Comment (k2d222) [inline Visibility.md:1]: editorial: would be good for readers to clarify early that app code needn't make anything public, since package-visibility from the root module suffice.

Comment (k2d222) [reply Visibility.md:1]: we could explore adding a linker option that turns all package declarations into public at publish time, for those who prefer the public-by-default mindset. Doesn't have to be a spec thing.

File: Visibility.md [hunk 30]
```diff
@@ -1,104 +1,255 @@
...
+* `public`: visible from any package.
+* *package* (the default): visible from any module in the same package.
+* `private`: visible only within the declaring module.
+
+| Visibility | Cross-module access | [Re-exportable](#a-re-export-cannot-widen-visibility) | Root-declared [pipeline-relevant item](#pipeline-visibility) |
+| --- | --- | --- | --- |
+| `public` | Any package | Yes | Pipeline-visible |
+| *package* (default) | Same package only | No | Pipeline-visible |
+| `private` | Declaring module only | No | Not pipeline-visible |
+
```
k2d222, line 18: here I would add a "note: in app code, one doesn't need to write any public because package suffice to be pipeline-relevant item."

Visibility.md
```diff
@@ -1,104 +1,255 @@
...
+## Visibility between modules
+
+A module can use an item from another module (named through `import`, or through
+an inline path such as `other_pkg::math::dot2`) only if that item is visible to
+it. A wildcard import omits items that are not visible to the importer. A name
+collision between two wildcard imports is therefore an error only when both
+items are visible (see [Wildcard imports](Imports.md#wildcard-imports)).
+
```
k2d222, line 26: we should also reference that fact in wildcard-imports

Visibility.md
```diff
@@ -1,104 +1,255 @@
...
+visible its referrers must be. A `public` declaration may refer to a less-visible
+one: a `public fn` can return or accept a *package* struct, and a `public struct`
+can have a field of a less-visible type. A consumer that cannot see the type can
+still use a value of it (call the function, read its fields) but cannot name the
+type to declare a variable, import it, or construct a new value.
+
+Naming a less-visible type in a `public` signature or field is permitted with no
+diagnostic.
+
+## Re-exports
```
k2d222, line 123: I would add: with re-exports, it's possible that a same declaration is reachable from different paths.

k2d222, line 123: can you re-export from "parent" modules? sounds like an anti-pattern, but I'm not sure.

k2d222 (changes requested):
Looking good! a bunch of editorial and nitpicks comments, otherwise there is just one thing I want to defend: re-exporting package-visible declarations, see the comment about Lygia.

Also, make it clear to readers that app users need not to worry about public it's a good selling point of the design!

Visibility.md
experimentslee: > I think it'd be good to allow re-exports of package declarations. This is really useful when you don't want to leak implementation details, e.g. you want your public-facing API to be all in one module, but internally you have implemented things in sub-modules.
> 
> I don't want to `import lygia::math::mod::mod`. I want `import lygia::math::mod`. 

I want that too. And we can have the good path in the current draft, we just can't hide the less favored one.

I like the part in the current design where visibility controls are fully explained at the declaration site. You know right away if you're refactoring a public api, w/o searching the project for re-exports.

Some ideas for hiding the less favored path:

- [@canonical](https://github.com/webgpu-tools/wesl-spec/blob/visibility/VisibilityDesign.md#canonical-prelude-path), an IDE hint w/o changing visibility rules.

- Mark something at the declaration site, `@hidden public fn mod() {}`. 

- Module or module path visibility controls, in addition to the item level controls. e.g. `@!public;` in the module. But that still makes the declaration site contract less clear.

I lean towards `@hidden public fn mod() {}` as I write this. But I'd propose we defer deciding, and not change the draft spec yet. Hiding the less favored path is a good feature, but extra polish. And there are some ideas floating around in related areas that might impact the choice (wesl.toml, module parameters, etc.). I'd add something to [Future possibilities](https://github.com/webgpu-tools/wesl-spec/blob/visibility/VisibilityDesign.md#future-possibilities) for now. wdyt?

Visibility.md
experimentslee: > I think an override should always be public, because it never makes sense otherwise. If it's non-public and initialized, it should just be a const. If it's non-public and non-initialized, it must be unreachable, otherwise it's a link error.
>
> The same idea goes for bindings and entrypoints. They should at the very least not be private.

package visible overrides, bindings and entry points declared in the root module are visible from host code. They don't need to be public in the current draft.

If they're in another module and you want to re-export them, they'd need to be public.

I agree private on these makes no sense.

>
> ... Or is this a warning better suited to a linter/language server? wdyt?

Single file WGSL shaders should just work in all their un-visibility-modified vanillaness.

So I think that means:
- Error on private overrides, bindings and entry points. You added `private`. You broke it, we can tell you.
- In libraries, warn on package visible overrides, bindings and entry points at publish time or if the lang server can distinguish library from app.
- In apps, outside the root module, optionally warn on package visible overrides, bindings and entry points. Optional because we might fix this with `package import`, allowing a root module to reexport a package visible override in an app. Optional also because we don't have a reliable way for the lang server to identify root modules.

Imports.md
mighdoll: I'd vote no, but I think we should clarify the spec. (in a later PR)

Visibility.md
mighdoll: Nothing stops it, agree seems likely bad style, good for a linter rule.

Visibility.md
mighdoll: good catch, will fix. we shouldn't make invalid code valid

Visibility.md
k2d222: I like re-exports better than the hidden annotation. Hidden sounds like extra steps to me. Pub at the call site explains your intent to publish, but it simultaneously hides your intent to _not publish here and instead_ move it to another module. The re-export solution is cleaner imo, it tells both _what_ to publish and _where_.

From my perspective it's not breaking any contract, because the declaring module already made the declaration visible to other modules, and what these modules do with the declaration is their business. I mean, they could just publish a wrapper function, but it's not a pattern we want to encourage.

k2d222: Alright, in the spirit of getting things moving faster, I agree to merging now, but I haven't changed my mind! We need to find a solution for this problem rather soon than later. I propose we clean up and merge this PR, and open a separate issue for "module visibility".

k2d222:
Cf. https://github.com/webgpu-tools/wesl-spec/issues/202#issuecomment-4990849899, We should treat this case in the spec:

```wgsl
private struct foo { ... }
public alias bar = foo; // is bar public? can I access its fields?
```

I think i this case `bar` should be public but be an opaque type.

mighdoll:
> I think in this case `bar` should be public but be an opaque type.

Good one. I agree we'll want something so that APIs can return structs with hidden internal details. I think it deserves its own issue. Filed #204.

> ```wgsl
> private struct foo { ... }
> public alias bar = foo; // is bar public? can I access its fields?
> ```

I'll update to make the alias an error in this PR while we consider what we want in #204. WGSL alias semantics say that aliases don't create new types. Should they create new visibility? Maybe, but there are other approaches to opaqueness that I suspect we'll like better..

Visibility.md
mighdoll: tried a fix in the latest draft (tho different wording w/o the private analogy)

Visibility.md
mighdoll: done

Visibility.md
mighdoll: revised in latest draft

Visibility.md
mighdoll: revised in latest draft

Visibility.md
mighdoll: revised

Visibility.md
mighdoll: revised

Visibility.md
mighdoll: revised

Visibility.md
mighdoll: > editorial: would be good for readers to clarify early that app code needn't make anything public, since package-visibility from the root module suffice.

revised to include this

GLOSSARY.md
mighdoll: later sounds good... (I tried this but realized that distinguishing 'errors' from 'link errors' is probably a bigger/separate issue than just a glossary entry)

Visibility.md
mighdoll: I haven't yet applied this one. Will do soon, though let me know if you've further thoughts.

Visibility.md
k2d222: filed #202

Visibility.md
```diff
@@ -75,21 +87,53 @@ can have a field of a less-visible type. A consumer that cannot see the type can
...
+
+An alias whose target resolves through any chain of aliases to a declared
+structure type must not be more visible than the structure declaration. A WGSL
+alias introduces another name for the same type and preserves its value
+constructors and members. A more-visible alias would therefore make the type
+nameable and constructible outside its visibility boundary (see
+[A re-export cannot widen visibility](#a-re-export-cannot-widen-visibility)).
+The visibility of an intermediate alias does not constrain a later alias. For
+example, `public alias Shared = Local;` is permitted when `Local` is a
+*package*-visible alias for a `public` structure type.
```
k2d222, line 117: I'm not 100% sure this is the best solution here, but we can keep it this way for now. In -rs, the 1st implementation iteration will probably allow public aliases to private structs, because checking the spec rule is harder.

mighdoll: sounds like a good plan. It's a corner case that few users will notice. #202 and #204 should give us some more clarity on visibility widening and opaque views of types, and then we can then adjust the alias rule and the fn signature rules about less visible types consistently.

k2d222 (approved):

Visibility.md
mighdoll: revised in latest draft:

+Marking an entry point or resource variable `private` is an error: `private`
+items are never pipeline-visible, so the host could never select or bind them.
+An `override` may be `private`: it is not host-settable, but it can still
+participate in the shader by deriving its value from another override (see
+[Pipeline-overridable constants](#pipeline-overridable-constants)).

stefnotch (commented):
I found an edge case that we'll have to test.

```
// in foo.wesl
public import package::utils as GetUtils; // a module

// in package.wesl
import package::foo::GetUtils;

const a = GetUtils::SOME_CONST;
```

If I do it correctly, I first expand to
```
const a = package::foo::GetUtils::SOME_CONST;
```
and then I resolve.

If I do it as a two step process, I might do the incorrect
```
// resolves to package::utils
import package::foo::GetUtils;

// so this resolves to 
const a = package::utils::SOME_CONST;
```

We should add that to the testsuite. And I'll have to be careful in wgsl-analyzer. W-a currently purposefully tries to do it as a two step process so that we can detect errors earlier.

Visibility.md
```diff
@@ -1,104 +1,310 @@
...
+
+@fragment
+fn fragment_main() -> @location(0) vec4f {
+  return apply_lighting();
+}
+
+// sun_intensity not in the pipeline-visible API: host cannot set it; uses its default of 1.0
+```
+
+### Aggregating entry points
```
stefnotch, line 293: I wonder if we could design an even neater world if we start messing with the host APIs. Like "any entrypoint in any module that is `public` can be used, the host API just has to say which one it wants"

stefnotch:
I approve of shipping this and implementing it. There's no reason to block further progress on small comments. 

If we come up with significantly better solutions, we can always ship a new iteration of the spec.

PR #188, opened 2026-05-12
client side builtin_shadow diagnostic

mighdoll:
builtin_shadow diagnostic warns clients if they're wildcard overriding a builtin (supressible)

### DIFF
--- a/Imports.md
+++ b/Imports.md
@@ -336,7 +336,9 @@ expected semantics that oughtn't be implicitly overridden with wildcards.
 Similarly, avoid experimental Naga/Dawn/Safari builtins.
 - WESL publishing tools should warn when a `@wildcardable` module exports an
   item that shadows a WGSL builtin. Suppress with
-  `@diagnostic(off, builtin_shadow)` in the module if the shadow is intentional.
+  `@diagnostic(off, builtin_shadow)` if the shadow is intentional.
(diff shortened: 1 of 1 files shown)


stefnotch:
Are there other cases where we should warn when a builtin in shadowed?

For example `alias f32 = i32;` should definitely result in some diagnostic.

Imports.md
```diff
@@ -399,13 +402,13 @@ The fix is to disambiguate with a named import (`import foo::clashing_zap;`) or
...
-  `@diagnostic(off, wildcard_shadow)` on the shadowing declaration or import.
+  [Scope precedence](#scope-precedence)).
 
-- **`builtin_shadow`** fires in WESL publishing tools when a `@wildcardable`
-  module exports an item that shadows a WGSL builtin such as `vec3` or `clamp`.
-  Suppress with `@diagnostic(off, builtin_shadow)` in the module if the override
-  is intentional.
+- **`builtin_shadow`** fires on a wildcard import when a referenced name in the
+  module resolves to a wildcard-imported item that shadows a WGSL builtin such
+  as `vec3` or `clamp`. Suppress at the import site if the override is
```
k2d222, line 409: just import site or both declaration site and import site? I understood it's both from a sentence above.

Issue #191, opened 2026-06-12
Restrict grammar for @if


stefnotch:
The wesl spec allows for attribute before assignments
```wesl
@if(foo)
a = 3;
```

However, I am not sure how to implement that in the language server's parser. It requires arbitrary, complex lookahead. As in, the following is a problem to parse
```
@some_attr
(very + complex * (expression)) = 3;
```

When the parser sees the `(`, it has to make a choice. Does it parse it as an attribute `@some_attr(very + complex * (expression))` or is the attribute finished `(very + complex * (expression)) = 3;`. With backtracking, that's doable. I really do not want backtracking in the language server parser, as it
- is significantly more complex to maintain
  - would mess with our error recovery capabilities
- introduces a source of slowdowns even for just WGSL (backtracking is a slowdown)
- still does not address the fundamental issue of this grammar being ambiguous

To elaborate on the last point, the WGSL grammar recently got extended to allow for function calls in the [lhs side of an expression](https://www.w3.org/TR/WGSL/#syntax-lhs_expression). The following is now *syntactically* valid WGSL.
```wgsl
foo() = 3;
(foo()) = 3;
```

If WGSL or WESL ever gets function pointers, then this would be valid WGSL. `foo.bar` is the function pointer and it's called with `value`
```wgsl
(foo.bar)(value) = 3;
```

Combining it with an ambiguous attribute, we get
```wgsl
@if
(foo.bar)(value) = 3;
```
Which is completely ambiguous. `(foo.bar)` might either be a part of the attribute or not. We can't have ambiguous grammars.

Now we could say that `@if` always requires `(` after it and that parsing that attribute needs to happen during the parsing step, instead of afterwards.
[...]

PR #196, opened 2026-06-26
More precise grammar for conditional translation

stefnotch:
Conditional translation in the spec and in the implementations does not fully match up. Time to fix this. 

### Narrower grammar for `@if`

https://github.com/webgpu-tools/wesl-spec/issues/191#issuecomment-4700814873

Both in wesl-rs and wgsl-analyzer, the current grammar of `@if` has been causing implementation woes. To address this with the smallest spec change, I opted to separate "attributes" and "translate-time attributes". Translate-time attributes are `@if(...)`, `@elif(...)` and `@else`. They are parsed exactly as such, thus being unambiguous. This change does not reject any existing WESL programs.

### `@else` in the spec

https://github.com/webgpu-tools/wesl-spec/issues/117

Both wesl-js and wesl-rs support `@elif` and `@else`. This adds them to the spec 

Compared to https://github.com/webgpu-tools/wesl-spec/pull/164 , I did not yet update the name to "condition attributes". I wanted to keep the diff easy to review for this PR.

### DIFF
--- a/ConditionalTranslation.md
+++ b/ConditionalTranslation.md
@@ -123,7 +123,7 @@ Checking for multiple features is done with an `&&`
 @if(feature1 && feature2)   const decl: u32 = 0;
 ```
 
-- `@if` takes a single parameter.
+- An `@if` attribute takes a single parameter. It marks the decorated node for removal if the parameter evaluates to `false`.
 
 - An `@elif` attribute decorates the next sibling of a syntax node decorated by a `@if` or an `@elif`. It takes one parameter.
 It marks the decorated node for removal if its parameter evaluates to `false` OR if any of the previous `@if` and `@elif` attribute parameters evaluate to `true`.
--- a/ConditionalTranslation.md
+++ b/ConditionalTranslation.md
@@ -113,34 +113,39 @@ Refer to the [updated grammar appendix](#appendix-updated-grammar) for the list 
    * function call statements
    * const assertion statements
 
-> [!WARNING]
-> We added attributes on assignment and increment/decrement statements. This change makes the WESL grammar no longer [LR(1)](https://en.wikipedia.org/wiki/LR_parser). See [`wesl-rs#162`](https://github.com/webgpu-tools/wesl-rs/issues/162) for details.
->
-> Due to this limitation, *wesl-rs* does not allow attributes on assignment, increment and decrement statements starting with a `(`. Concretely, this is not supported by *wesl-rs*: `@if(FOO) (x)++`.
+## `@if` attribute family
+The `@if`, `@elif` and `@else` *translate-time attributes* are introduced. The decorated node is only kept if the branch is truthy.
 
-## `@if` attribute
-The `@if` *translate-time attribute* is introduced. It takes a single parameter. It marks the decorated node for removal if the parameter evaluates to `false`.
-
-A syntax node may at most have a single `@if` attribute. This keeps the way open for a `@else` attribute introduction in the future.
+A syntax node may at most have a single `@if`, `@elif` or `@else` attribute. 
 Checking for multiple features is done with an `&&`
 
 ```wgsl
 @if(feature1 && feature2)   const decl: u32 = 0;
 ```
 
-> *See the [possible future extensions](#possible-future-extensions) for the attributes `@elif` and `@else`.
-> They may be introduced in the specification in a future version if deemed useful.*
+- `@if` takes a single parameter.
+
+- An `@elif` attribute decorates the next sibling of a syntax node decorated by a `@if` or an `@elif`. It takes one parameter.
+It marks the decorated node for removal if its parameter evaluates to `false` OR if any of the previous `@if` and `@elif` attribute parameters evaluate to `true`.
+
+- An `@else` attribute decorates the next sibling of a syntax node decorated by a `@if` or an `@elif`. It does not take any parameter.
+It marks the decorated node for removal if any of the previous `@if` and `@elif` attribute parameters evaluate to `true`.
 
 Example:
 
 ```wgsl
 @if(feature_1 && (!feature_2 || feature_3))
 fn f() { ... }
-@if(!feature_1)                               // corresponds to @elif(!feature_1)
+@elif(!feature_1)
 fn f() { ... }
-@if(feature_1 && !(!feature_2 || feature_3))  // corresponds to @else
+@else
 fn f() { ... }
 ```
+
+
+
+Design note:
+To keep the grammar [LR(1)](https://en.wikipedia.org/wiki/LR_parser), all translate time attributes are unambiguously specified in the formal grammar. This allows for translate time attributes before assignment, increment and decrement statements starting with a `(`. A concrete example is `@if(FOO) (x)++`.
 
 ## Execution of the conditional translation phase
 1. The *WESL translator* is invoked with the list of features to *enable* or *disable*.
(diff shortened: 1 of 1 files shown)


ConditionalTranslation.md
```diff
@@ -113,35 +113,40 @@ Refer to the [updated grammar appendix](#appendix-updated-grammar) for the list
...
 fn f() { ... }
-@if(feature_1 && !(!feature_2 || feature_3))  // corresponds to @else
+@else
 fn f() { ... }
 ```
 
+
+
+Design note:
+To keep the grammar [LR(1)](https://en.wikipedia.org/wiki/LR_parser), all translate time attributes are unambiguously specified in the formal grammar. This allows for translate time attributes before assignment, increment and decrement statements starting with a `(`. A concrete example is `@if(FOO) (x)++`.
```
k2d222: clarify what unambiguously specified means. I think you mean: t-t attributes are a distinct prodution rule from regular attributes, because regular attrs would be ambiguous in certain locations. Concrete example...

ConditionalTranslation.md
```diff
@@ -113,35 +113,40 @@ Refer to the [updated grammar appendix](#appendix-updated-grammar) for the list
...
+A syntax node may at most have a single `@if`, `@elif` or `@else` attribute. 
 Checking for multiple features is done with an `&&`
 
 ```wgsl
 @if(feature1 && feature2)   const decl: u32 = 0;
 ```
 
-> *See the [possible future extensions](#possible-future-extensions) for the attributes `@elif` and `@else`.
-> They may be introduced in the specification in a future version if deemed useful.*
+- `@if` takes a single parameter.
```
k2d222: here you dont say what if does: "An `@if` attribute takes one parameter. It marks the decorated node for removal if the parameter evaluates to `false`."

ConditionalTranslation.md
k2d222: you can also make this a note with `[!NOTE]`

stefnotch: I'm not sure how to phrase this in a succinct way. https://github.com/webgpu-tools/wesl-spec/issues/191 is kind of long.

k2d222: I mean it's not at all obvious that the grammar is unambiguous here, and I genuinely wasn't sure if/why it was at first. The root of the ambiguity is the _optional_ `(` in the `attribute` production rule, which is ambiguous with statements starting with `(`. It would not be ambiguous if `(` was mandatory or forbidden. 

But your solution introduces a context-sensitive token: 'elif'. Before this change, one could just parse all attributes as `'@' Ident`, now the lexer has to track if the previous token is `@`. No big deal, but we should mention it.

But if we use context-sensitive lexing, then why not make *all* attribute names context-sensitive, remove the catch-all rule (`@ Ident OptionalParen`), and then the `(` is no longer optional, and the ambiguity disappears! We don't need the separate `translate_time_attribute` rule, Hooray! ...But that only works because each attribute has either no paren or a mandatory paren, so it's a bit fragile if we do introduce one in the future.

But actually it can work *without* context-sensitive lexing, if you rewrite `translate_time_attribute` as `'@' 'if' Paren | '@' 'else' | '@' Ident Paren`, making the `(` non-optional. But then, the `attribute` must become just `translate_time_attribute | '@' Ident`, because the `'@' Ident Paren` is already taken. And the name `translate_time_attribute` is no longer accurate.

**TL;DR**: idk what the best solution is here, but these are the options:
* keep syntax as-is, but tell that `elif` is context-sensitive, and that `translate-time-attribute` is a sub-rule of `attribute`.
* make all attributes context-sensitive, and have just one `attribute` rule everywhere.
* make attributes non-context-sensitive, and rewrite/rename the two-stage ambiguous/unambiguous attribute rules.

stefnotch: Ah, I had missed that. I chose your final option without context sensitivity. I did it with one extra rule, hopefully that's reasonably clear.

(Although I'm not absolutely sure if that really solves it.)

k2d222: I should have added 1 more detail: the WGSL grammar already says that attribute names are context-sensitive. But it implicitly leaves room for implementations to be non-context-sensitive, by adding a catch-all rule for "unknown" attributes.

Your commit does not solve it, here's how to solve it:

* You want `elif` to not be context-sensitive, so you must have `'@' ident_pattern_token argument_expression_list` in the 'translate_time_attribute' rule.
* You want `attribute` to contain `translate_time_attribute`, but they must not conflict (you can't allow a syntax form in `attribute` that's already in `translate_time_attribute`), so you must replace `'@' ident_pattern_token argument_expression_list ?` with just `'@' ident_pattern_token`

solution:
```
attribute:
    // the rule `'@' ident_pattern_token argument_expression_list` is already in `unambiguous_attribute`, having it here too would be a Reduce-Reduce conflict.
    | unambiguous_attribute
    | '@' ident_pattern_token
    | ...

unambiguous_attribute:
    | '@' ident_pattern_token argument_expression_list
    // the WGSL grammar for attributes contains both a catch-all production rule, and more specific rules *if you support* context-sensitive tokens. If your lexer *does not support* them, you can just skip the rules below.
    | '@' 'if' '(' expression ',' ? ')'
    | '@' 'elif' '(' expression ',' ? ')'
    | '@' 'else'
```

stefnotch: Wait, I don't get how that is different?

`unambiguous_attribute` corresponds to `translate_time_attribute` (if you ignore the indirection)

And `attribute` also expands to the same set of rules. It has a `'@' ident_pattern_token argument_expression_list ?`

k2d222: for a bottom-up LR parser, you can't have two child rules that accept the same syntax phrase.

In your commit, you have `translate_time_attribute` which accepts `'@' Ident argument_expression_list`, and `attribute` which accepts `'@' Ident argument_expression_list ?`. The latter has the extra `?`, but they both accept "@attr(expr)".

the pedantic explanation: when a parser is in a state where it has read `@ ident argument_expression_list`, and can reduce it to an `attribute`, then necessarily it can also reduce it to a `translate_time_attribute`, because the latter is a sub-rule of `attribute`, so it must be available at that point. It doesn't know whether to do `@ ident argument_expression_list` -> `attribute`, or `@ ident argument_expression_list` -> `translate_time_attribute` -> `attribute`.

The smallest fix from your PR is simply changing `'@' Ident argument_expression_list ?` to `'@' Ident` in `attribute`, but notice that it also changes the semantics of `translate_time_attribute`, which now also captures all attributes with parenthesis. That's why I renamed it.

stefnotch: Oh I understand. If I see a `@`, I do not yet know whether I'll parse a `translate_time_attribute` or a normal attribute. So the rules do conflict.

Then I guess I'll apply your fix. 

As for the alternative of context sensitive lexing: It adds so much baggage to the lexers that I prefer to avoid it.

k2d222: yeah, and the WGSL spec is a tiny bit vague here. They say attributes are context-sensitive, but they do provide this catch-all grammar rule which serve no purpose, whithout explaining that it's an escape hatch for non-context-sensitive lexers.

ConditionalTranslation.md
```diff
@@ -257,77 +212,91 @@ The following non-terminals are added or modified:
...
-
-    if_attr:
-      '@' 'if' '(' expression ',' ? ')'
+    | ...
+    | known_translate_time_attribute
 ```
+
+### Possible extensions
+
+The `@else` attribute has the nice property that all cases lead to generating a node, and *could* therefore be used in places where the node is required (e.g. expressions)
```
mighdoll: restore bullet

ConditionalTranslation.md
```diff
@@ -113,35 +113,40 @@ Refer to the [updated grammar appendix](#appendix-updated-grammar) for the list
...
 fn f() { ... }
-@if(feature_1 && !(!feature_2 || feature_3))  // corresponds to @else
+@else
 fn f() { ... }
 ```
 
+
+
+Design note:
+To keep the grammar [LR(1)](https://en.wikipedia.org/wiki/LR_parser), all translate time attributes are unambiguously specified in the formal grammar. This allows for translate time attributes before assignment, increment and decrement statements starting with a `(`. A concrete example is `@if(FOO) (x)++`.
```
mighdoll: perhaps something like:

[!NOTE]
No alternative of `unambiguous_attribute` has an *optional* argument list, so a parser always knows where the attribute ends. `attribute` lets the argument list be omitted, which is ambiguous before a statement: in `@foo (x)++`, the `(x)` could be `@foo`'s argument or the start of the statement. Using `unambiguous_attribute` keeps the grammar [LR(1)](https://en.wikipedia.org/wiki/LR_parser) and allows`@if(FOO) (x)++`.

mighdoll (commented):
looks good, editorial suggestions here.

ConditionalTranslation.md
mighdoll: `unambiguous_attribute` looks fine to me. The comment would only apply to `elif` tho right?  `if` and `else` are keywords, which WGSL says can't be attribute names. So they'd need their own rules regardless.

ConditionalTranslation.md
stefnotch: Thank you! That reads nicely.

ConditionalTranslation.md
stefnotch: Okay, I pushed the new version. This grammar tweak feels a big ugly, but I suppose it's the best we can do for now.

k2d222 (approved):

mighdoll (approved):

Issue #193, opened 2026-06-22
Imports corresponding to file paths


stefnotch:
### Background

Both @mighdoll  and @k2d222  observed that as long as there are no module re-exports, one can skip the entire import resolution algorithm and directly go to the files. I originally was going to [post this on Discord, but the message became very long](https://discord.com/channels/1275293995152703488/1518533028794601492/1518533034196996208).

## Exploration

Assuming a world where `package::foo::bar()` really does mean `import("./shaders/foo.wesl").bar()`, then we get the following behaviours:

`import package::foo::bar` is *ambiguous until usage*. It could either refer to an item (`const bar = import("./shaders/foo.wesl").bar;`) or it could refer to the entire module (`const bar = import("./shaders/foo/bar.wesl");`). When it gets used, like `bar()` or `bar::nya()`, then it is clear. *This behaviour is fine, because our imports are lazy.* 

However, it does mean that `::` very much corresponds to a file path. Associated constants and associated types will have to use the `MyType.constant` syntax instead of `MyType::constant`. Enums will have to use `MyEnum.Variant` instead of `MyEnum::Variant`. Otherwise `package::MyType::constant` would be ambiguous with no way of telling what it is. 

### On the topic of re-exports

Re-exports could only be done at the root (e.g. in `wesl.toml`), because all the other files might never be imported. Bevy likely wants module re-exports at the top level. As in `bevy::pbr` being an alias for `bevy_pbr`. Re-exports deeper inside a shader module don't have any concrete use cases at the moment, I believe. We can later add module level re-exports, but it will be less ergonomic in this world.

### Visibility is simpler

Marking *modules* private is no longer a reasonable operation. I can always write `package::foo::internal::superhidden::private()` and access it, because it's a direct filesystem access. (`import("./shaders/foo/internal/superhidden.wesl").private()`) In my opinion, this would address https://github.com/webgpu-tools/wesl-spec/issues/103 .

So, the rule for modules should change to: A *module* is only as visible as its most visible item. If `superhidden.wesl` only has private items, then the module is private. This barely matters, module-level visibility only pops up in two places now
- Language server autocompletions for `package::foo::internal::|`. Private modules wouldn't show up.
- Re-exports, if we add a mechanism for them.

### Directly jumping to paths is nice

[...]

Issue #189, opened 2026-05-27
Bevy has multiple crates


stefnotch:
I have been thinking about the Bevy case a bit more:

So the way Bevy expects users to install it is by [adding a dependency on `bevy`](https://bevy.org/learn/quick-start/getting-started/#installing-bevy)
```toml
# Cargo.toml
[dependencies]
bevy = "0.18.1"
```

Meanwhile, Bevy itself is split into quite a few crates. Notable ones are `bevy_pbr` and `bevy_render`.
[For example, this](https://github.com/bevyengine/bevy/blob/564eb4fc3383b7b61d871ce1c28318f79459c56c/examples/shader/shader_material_bindless.rs#L7) imports [code from both](https://github.com/bevyengine/bevy/blob/main/assets/shaders/bindless_material.wgsl). There are also [other examples like [`bevy_sprite`](https://github.com/bevyengine/bevy/blob/main/assets/shaders/custom_material_2d.wgsl

For Bevy to adopt WESL, we actually do need some way of covering that use case. One that also works with the IDE.

With that in mind, here are some possible proposals. I currently only see "1. Module re-exports" as actually being satisfactory.

## 1. Module re-exports

This approach is stolen from Rust.
WESL users would add the Bevy dependency

```toml
# wesl.toml
[dependencies]
bevy = {}
```
Adding a dependency on `bevy_pbr` is not possible here. After all, `bevy_pbr` is an *indirect* dependency.[^1]

Bevy would ship with a shader that re-exports its dependencies. Note that here we are re-exporting entire modules.
```wgsl
// package.wesl
public import bevy_pbr as pbr;
public import bevy_render as render;
```

And finally, the user can write
```wgsl
// bindless_material.wgsl
import bevy::pbr::mesh_bindings::mesh;
import bevy::render::bindless::{bindless_samplers_filtering, bindless_textures_2d};
```

[^1]: Technically it would be possible, but semantically it'd be wrong. It would be an equivalent of the flat `node_modules` crime that npm does and pnpm tries to prevent.

## 2. Require dependencies on `bevy_pbr` and `bevy_render`

Bevy would have to tell their users to do
```toml
# Cargo.toml
[dependencies]
bevy = "0.18.1"
# Better keep these in sync ;)
bevy_pbr = "0.18.1"
bevy_render = "0.18.1"
```

And then the user can do

```toml
# wesl.toml
[dependencies]
bevy = {}
bevy_pbr = {}
[...]

PR #199, opened 2026-07-06
direct import resolution

mighdoll:
fix for #193

replace the left-to-right segment walk with direct resolution:
the last path segment names the declared item, the preceding segments name the module.
    
declarations and modules naturally live in separate namespaces now
- `fn foo() {}` and `foo::bar();` do not conflict
    
lazy vs eager resolution is an implementation choice
    
clarify terms a bit

### DIFF
--- a/Imports.md
+++ b/Imports.md
@@ -131,7 +131,6 @@ The absolute path of the `super` module is always known, since the first loaded 
 Once the import has been resolved, the last segment, or its alias, is brought into scope.
 
 The order of the scopes is "user declarations and imported items > wildcard import > package names > predeclared items".
-This lets WGSL add more predeclared items without breaking existing WESL code. Package names can shadow predeclared items, but we recommend that authors avoid doing that.
 
 
 ### Example
@@ -266,17 +265,34 @@ import bar::*; // exports clashing_name
 The name clash is ignored until the item is used.
 TODO: Or should it immediately result in a warning, even if the variables are unused?
 
-### Wildcard imports from libraries 
+### Diagnostics for best practices
 
-For libraries, we distinguish between modules that were designed for wildcard imports and modules that were not.
-Library authors generally expect that adding a new item is not a breaking change. Wildcard imports are a surprising edge case.
+Many ecosystems rely on semantic versioning for updating libraries. This applies to both direct and transitive dependencies. Updating to the latest patch releases should be easy and safe. 
 
-WESL detects the edge cases and will emit a diagnostic. This makes upholding semver guarantees significantly easier. 
+Library authors generally expect that adding a new item is not a breaking change. Wildcard imports that result in name clashes are a surprising edge case. Therefore, we want to reduce the risk of accidental name clashes.
 
-When the library module is called `prelude`, it is a module designed for wildcards. Adding a new item to it is a semver breaking change.
-Wildcard importing from it is always allowed. (When we add exports, we will make this controllable instead of only giving special treatment to the `prelude`.)
+Name clashes can happen either in user code or in library code. If it happens in library code, it can be significantly harder to resolve the issue. 
 
-Other library modules are not designed for wildcards. If another wildcard import exists, then an `wildcard_import` [diagnostic](https://www.w3.org/TR/WGSL/#diagnostics) at the error level will be emitted.
+A secondary, similar issue is that wildcards can result in accidental shadowing of predeclared identifiers. 
+One especially insideous scenario is when user code targets an unstable Naga feature.
+
+```wesl
+import my_lib::*;
+
+const x = unstable_thing();
+```
+
+Then, a library which does not know about the Naga feature could add an item with the same name. 
+
+To reduce the odds of these cases, we recommend the following diagnostics.
+
+1. Emit a warning [diagnostic](https://www.w3.org/TR/WGSL/#diagnostics) named `builtin_shadowing` if an top level item is declared that is known to conflict with a predeclared identifier. This also helps in the non-library case.
+
+2. Emit a warning [diagnostic](https://www.w3.org/TR/WGSL/#diagnostics) named `wildcard_import` if there are multiple wildcard imports from libraries and at least one is not designed for wildcards.
+
+For this, we distinguish between modules that were designed for wildcard imports and modules that were not. When the library module is called `prelude`, it is a module designed for wildcards. Adding a new item to it is a semver breaking change. Therefore, wildcard importing from it is always allowed. (When we add exports, we will make this controllable instead of only giving special treatment to the `prelude`.)
+
+Other library modules may cause issues and should not be combined with other wildcard imports.
 
 Modules in the same package can always be wildcard imported. Changing your own code has no impact on semver.
 
@@ -298,6 +314,18 @@ import package::utils::*;
 @diagnostic(off, wildcard_import) 
 import lygia::math::*; 
 ```
+
+### Modeling the WGSL predeclared identifiers
+
+One mental model for WGSL predeclared identifiers is that they're a wildcard import, which is implicitly added to every WGSL file. 
+
+```wesl
+// imports all the predeclared identifiers
+import all_wgsl_items::*;
+```
+
+This wildcard import has a strictly lower priority than all other wildcard imports. This lets WGSL add more predeclared items without breaking existing WESL code. Package names can also shadow predeclared items, but we recommend that authors avoid doing that.
+
 
 ## Directives
 Under discussion, see: <https://github.com/wgsl-tooling-wg/wesl-spec/issues/71>
--- a/Imports.md
+++ b/Imports.md
@@ -22,6 +22,9 @@ import my::geom::sphere::{ draw, default_radius as foobar };
 
 // Imports a whole module. Use it with `bevy_ui::name`
 import bevy_ui;
+
+// Import all items defined in the bevy prelude
+import bevy::prelude::*;
 ```
 
 These can then be used anywhere in the source code.
@@ -74,6 +77,7 @@ import_relative:
 import_path_or_item:
 | ident '::' (import_collection | import_path_or_item) 
 | ident ('as' ident)?
+| '*'
 
 import_collection:
 | '{' (import_path_or_item) (',' (import_path_or_item))* ','? '}'
@@ -90,6 +94,8 @@ Lint tools may optionally warn when reserved words are used.
 An item import imports a single item. The item can be renamed with the `as` keyword.
 
 An import collection imports multiple items, and allows for nested imports.
+
+A wildcard import imports all items from a module, but does not import further modules.
 
 ### Import resolution algorithm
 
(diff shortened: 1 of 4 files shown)


Imports.md
```diff
@@ -137,59 +139,129 @@ import foo::*;
...
+package served over the web can map each module path to a URL, and a bundled
+library can store module sources in a dictionary keyed by module path.
+
+A module path may consist of just `package`, or of just a bare package name.
+Such a path refers to the *package module*: on a filesystem,
+the file `package.wesl` in the package's root directory. The package module
+is optional.
+
+Referencing a declaration path that fails to resolve is an error. An import
+statement whose bound name is never referenced is allowed, even if
```
stefnotch, line 206: Seems reasonable, making import *statements* effectively lazy. It's the usage that must trigger the error. Although a language server can already warn beforehand :)

Imports.md
```diff
@@ -137,59 +139,129 @@ import foo::*;
...
+
+fn shade() {
+    let s = shadowmapping::pcf();
+}
 ```
 
-Assume that the current module lives at `shaders/lighting.wesl`. We first go to the super module at `shaders.wesl`. We then look for an item called `shadowmapping` in `shaders.wesl`.
-After not finding it, we look for a module `shadowmapping` at `shaders/shadowmapping.wesl`.
+The current module is stored at `lighting/pbr.wesl`, with module path
+`package::lighting::pbr`. `super` removes the last segment, giving
```
stefnotch, line 248: This is a good way of doing it. It nicely deals with super paths that refer to the package. Like `import super::super::something_from_the_package;`.
We would start with `package::lighting::pbr`, remove the segments and get `package::something_from_the_package`

stefnotch (approved):
Usually I'd bikeshed a lot, but I already independently implemented this.

This seems reasonable to me, from both a spec and an implementation standpoint. I approve of this.

Imports.md
```diff
@@ -102,61 +128,140 @@ import a::c::d;
...
+The first segment anchors the path:
+
+* `package` anchors the path at the current package.
+* `super` refers to the parent of the current module, removing the last
+  segment of the current module's path. Each additional `super` removes
+  another segment; it is an error to remove beyond the package root.
+* Any other first segment must name a known package, and anchors the path at
+  that package. Tools find the known packages in the
+  [`wesl.toml`](WeslToml.md) file or through the host package manager's
+  dependencies.
```
k2d222, line 186: is the current package name importable? I forgot what we agreed on.

Imports.md
```diff
@@ -102,61 +128,140 @@ import a::c::d;
...
+
+A package amounts to a mapping from module paths to module sources; the
+semantics of the module path segments beyond the first are specific to the
+package's storage. Only the module path maps to storage; the final segment of
```
k2d222, line 190: it's not clear who/what decides the storage strategy for a given package. Is it implementation-specific, declared by the package, implicit from usage, etc.

Imports.md
```diff
@@ -102,61 +128,140 @@ import a::c::d;
...
+a declaration path names a declaration inside the module source, never a file.
+For a package stored on a filesystem, the first segment refers to the
```
k2d222, line 192: Is this normative, i.e. do all filesystem storages have to behave this way?

Imports.md
```diff
@@ -102,61 +128,140 @@ import a::c::d;
...
+package's root directory; each following segment except the last names a
+subdirectory, and the last segment of the module path names the source file:
+`seg.wesl`, or `seg.wgsl` when no `.wesl` file exists
+(see [Filesystem Resolution](#filesystem-resolution)). A
+package served over the web can map each module path to a URL, and a bundled
```
k2d222, line 197: ... because here, it says "can", whereas for the filesystem storage, there is no conditional.

Imports.md
```diff
@@ -87,13 +90,36 @@ not current keywords are allowed,
...
+Attributes may precede an import statement, notably `@if` for
+[conditional translation](ConditionalTranslation.md) and `@diagnostic` for
+[suppressible diagnostics](#suppressible-diagnostics).
+
 An item import imports a single item. The item can be renamed with the `as` keyword.
 
 An import collection imports multiple items, and allows for nested imports.
 
-### Import resolution algorithm
+A wildcard import imports all top-level declarations from a module. Submodule names and submodule contents are not imported. A wildcard must follow a module path; a bare `import *;` is an error. A wildcard may also appear as a member of an import collection, applying to the module path before the braces (see [Import bindings](#import-bindings)).
```
k2d222, line 101: It looks like the direct import resolution no longer warrants that intermediate submodule exist, i.e `import lygia::math::abs` is a valid module path, but `import lygia::math` is not. Is that the case?

Imports.md
```diff
@@ -102,61 +128,140 @@ import a::c::d;
...
 
 ## `wesl.toml`
-The [`wesl.toml`](WeslToml.md) file provides linker configuration options affecting the import resolution algorithm. It can customize:
+The [`wesl.toml`](WeslToml.md) file provides linker configuration options affecting import resolution. It can customize:
 
 * The root directory,
 * Available package dependencies,
 * A file whitelist and/or blacklist.
 
 ## Filesystem Resolution
```
k2d222, line 262: see previous comment: let's say that the filesystem is a normative and the default storage mode, but other implementation-defined storages can exist.

Imports.md
```diff
@@ -102,61 +128,140 @@ import a::c::d;
...
-To resolve a module on a filesystem, one follows the algorithm above.
+To resolve a module on a filesystem, one follows the mapping in
+[Resolving a declaration path](#resolving-a-declaration-path).
 The root folder, or the root module, needs to be provided to the linker. This is currently a linker-specific API, and may change once we introduce a `wesl.toml`.
 
 Linkers should fall back to `.wgsl` files when a `.wesl` file cannot be found.
```
k2d222, line 267: must?

ImportsDesign.md
k2d222, line 1: can we move sec. rationale and alternatives here.

Imports.md
```diff
@@ -102,61 +128,140 @@ import a::c::d;
...
+statement whose bound name is never referenced is allowed, even if
+referencing it would be an error; tools may warn about unused or unresolvable
+imports. Import statements and references removed by
+[conditional translation](ConditionalTranslation.md) are also allowed, even
+if they would be errors under other conditions; tools may warn about these
+too.
+
+Tools can enumerate the potential resolutions an import statement
+enables by analyzing the source tree paths and the declarations in each
+module source, for example to suggest editor auto-completions.
```
k2d222, line 215: I don't think this paragraph is necessary here

Imports.md
```diff
@@ -102,61 +128,140 @@ import a::c::d;
...
+The [`wesl.toml`](WeslToml.md) file provides linker configuration options affecting import resolution. It can customize:
 
 * The root directory,
 * Available package dependencies,
 * A file whitelist and/or blacklist.
 
 ## Filesystem Resolution
-To resolve a module on a filesystem, one follows the algorithm above.
+To resolve a module on a filesystem, one follows the mapping in
+[Resolving a declaration path](#resolving-a-declaration-path).
```
k2d222, line 264: not just that: we start with a filesystem path prefix provided by the linker or wesl.toml, we append the segments from the mapping above, and then we suffix the last segment with .wesl or .wgsl. That's the complete fs path.

k2d222 (changes requested):
Two main comments:

* before, we had this condition:
  >  File not found: We assume an empty module as the current module, and continue with that.

  now idk what happens when importing an intermediate module path.

* It's unclear who decides to use the filesystem storage or some other storage. It's unclear that the fs storage is normative, but the others are implementation-defined.

Imports.md
stefnotch: There is no "current package name" in the spec at the moment. That should be a separate issue to deal with.
(Lygia wanted it after all.)

Maybe we'll do it with the same thing that was proposed here: https://github.com/webgpu-tools/wesl-spec/issues/189#issuecomment-4794591593 ?

Imports.md
stefnotch: Yes, exactly. Intermediate submodules are no longer necessary.

k2d222: if so then there are some wording tweaks to do. For instance: `import foo::bar::*` is only valid if `foo::bar` is a full path to a module. But `import foo::bar` is valid even if `foo::bar` is only an intermediate path (not a valid module), because then I can do `bar::baz::decl()`.

Imports.md
k2d222: or could be a note

ImportsDesign.md
mighdoll: I was sorely tempted, but seemed better as a separate PR.

Imports.md
mighdoll: I've rewritten this section on the visibility PR (as I was dealing with the merge conflicts). Take a look there? We can backport or perhaps just wait to merge that one.

Imports.md
mighdoll: Agreed, doesn't fit in the normative flow. Will try it as a note.

Imports.md
mighdoll: AFAIR we agreed it's importable. Fixing with #189 seems reasonable.

mighdoll:
> Two main comments:
> 
> * before, we had this condition:
>   > File not found: We assume an empty module as the current module, and continue with that.
>   
>   
>   now idk what happens when importing an intermediate module path.
> * It's unclear who decides to use the filesystem storage or some other storage. It's unclear that the fs storage is normative, but the others are implementation-defined.

See if it's more clear now on #199. 🤞

Imports.md
mighdoll: The fs storage mapping of module paths to directories is normative. 

I the latest #187 draft I tried writing the main algorithm w/o being storage system specific, so the rules are generic and the mapping is in the fs section

Imports.md
mighdoll: The package's storage decides. #187 now says:

> A package provides a mapping from module paths to module sources... How a
> package maps module paths to sources depends on how it is stored (see
> Filesystem Resolution and Non-Filesystem Resolution).

Imports.md
mighdoll: Good catch. The #187 separation of fs from the generic rule on module paths I think avoids this contradtion. and each storage has its own section, Filesystem Resolution says: ("a module path maps directly to a single file")

k2d222:
the conflicts between 199 and 197 make it hard to review. I say let't merge this and we can adjust the minute details in followup PRs.

stefnotch:
@k2d222 Am happy with that. Alternatively we could play around with rebasing.

stefnotch:
Okay, I tried out the rebase. Relatively straightforward with jj, but it would be a force push and it would affect the #187 pull request. Which, if I remember correctly, git does not like at all.

So I vote in favour of ~~everyone adopting jj~~ just merging this and fixing the issues afterwards.

Issue #222, opened 2026-08-19
Module parameters with < >


mighdoll:
Draft [proposal for module parameters](https://github.com/webgpu-tools/wesl-spec/blob/main/proposals/222-ModuleParameters.md) and some [design notes](https://github.com/webgpu-tools/wesl-spec/blob/main/proposals/222-ModuleParametersDesign.md). 

```wgsl
// lights.wesl - lights library, users can choose capacity
module <public const MAX: u32 = 8, public const DEBUG = false>;  // modules take parameters

var<private> list: array<Light, MAX>;     // params can size arrays
var<private> count: u32;

fn push(l: Light) {
  @if(DEBUG) { /* bounds check */ }       // conditions can use params
  list[count] = l;
  count += 1u;
}
```

```wgsl
import lights_lib::lights<MAX = 32>::{push, pop};  // apps can configure at import
```

## Why Module Parameters?

**Set module constants from host code** - Users want to inject constants from host code.

**Set module constants from shader code** - Libraries want to expose tunable parameters to users in other shader modules. For example, Lygia wants to let users set constants like `#define ATMOSPHERE_SUN_POWER 20.0` from [#132](https://github.com/webgpu-tools/wesl-spec/issues/132).

**@if conditions can use constants** - Conditions are currently restricted to host code set booleans, so e.g. bevy's WESL port encodes comparisons awkwardly in flag names like `@if(AVAILABLE_STORAGE_BUFFER_BINDINGS__GE_3)` in [#213](https://github.com/webgpu-tools/wesl-spec/issues/213).
 
## Related

- **Extends** [#146](https://github.com/webgpu-tools/wesl-spec/pull/146) and [#164](https://github.com/webgpu-tools/wesl-spec/pull/164) with revised syntax and some clarifications, and adapts capture semantics for types from zig/[#203](https://github.com/webgpu-tools/wesl-spec/issues/203).
- **Fixes**: [#132](https://github.com/webgpu-tools/wesl-spec/issues/132) (settable consts for e.g. Lygia), [#213](https://github.com/webgpu-tools/wesl-spec/issues/213) (numeric `@if` conditions for e.g. Bevy).
[...]

Issue #200, opened 2026-07-06
Dynamic pointers proposal


k2d222:
In https://github.com/webgpu-tools/wesl-spec/pull/147#issuecomment-4894356632 I propose using generalized pointers as an alternative to `@inline` functions, to fulfill the same goals, namely:

* passing textures and samplers as parameter to functions
* storing them in containers (struct or array)
* run-time or compile-time selection of textures / samplers
* polyfill of `unrestricted_pointer_parameters`

## Overview

* We introduce the `ptr<T>` type, called a dynamic pointer. In WGSL, regular pointers are spelled with a mandatory address space, e.g. `ptr<function, u32>`. Alternative syntax: `ptr<dyn, T>`.
* Any type can be stored in a `ptr<T>`. It can be copied, passed to functions and returned from them.
* a `ptr<T>` gets desugared to a `u32` which is an index into the list of possible instances the pointer may refer to.
* dereferencing a `ptr<T>` get desugared to a switch/case of the `u32` index.
* `ptr<T>` is opaque: the only allowed operations are declaration/assign/returning and dereference.

Rationale:
* Unlike `@inline`, the codegen is a lot more predictable: the function remains almost identical, and the branching happens at the dereference spot.
* Unlike `@inline`, it is capable of runtime branching, and thus enables the bindless pattern/paradigm very nicely.
* Unlike `@inline`, it allows storing textures/samplers in structs and passing them around.
* Like `@inline`, it can be zero-cost if the branching happens at shader-creation-time.
* I think it would be easier to implement than `@inline`.

## Example

```rs
@group(0) @binding(0) var myTex1: texture_2d<f32>;
@group(0) @binding(1) var myTex2: texture_2d<f32>;

fn select_texture(tex1: ptr<texture_2d<f32>>, tex2: ptr<texture_2d<f32>>, some_arg: f32) -> ptr<texture_2d<f32>> {
    if (some_arg > some_threshold) { return tex1; } else { return tex2; }
}

fn main() {
    let myTex: ptr<texture_2d<f32>> = select_texture(&myTex1, &myTex2, myCond);
    let texel = textureLoad(*tex, coords, 0);
}
```
The code above is rewritten as:
```rs
@group(0) @binding(0) var myTex1: texture_2d<f32>; // gets assigned ID 0
@group(0) @binding(1) var myTex2: texture_2d<f32>; // gets assigned ID 1

[...]

PR #223, opened 2026-08-19
module parameters with < > proposal

mighdoll:
module parameters proposal, goes with #222 

Please review this PR for form - is this how we'd like to do proposals? Approving this PR doesn't mean agreeing to the proposal, just that we have a proposal we can discuss.

I used Claude and Codex during the draft/review cycles though I wrote and rewrote almost every sentence. I think the AI helped overall - certainly it caught a number of inconsistencies and poorly written sentences in earlier drafts. 

I'm sure there'll be plenty to discuss on the actual module parameters topic, let's do that in #222!

### DIFF
--- a/proposals/222-ModuleParameters.md
+++ b/proposals/222-ModuleParameters.md
@@ -1,0 +1,123 @@
+# Module Parameters
+
+- **Status**: Proposal
+- **Discussion**: [wesl-spec#222](https://github.com/webgpu-tools/wesl-spec/issues/222)
+- **Extension marker**: `wesl_module_parameters`
+- **Implementations**: none
+- **Replaces / extends**: #146, #164; supersedes the unspec'd `constants::` virtual module
+
+See also [ModuleParametersDesign.md](ModuleParametersDesign.md) (design background).
+
+## Declaring module parameters
+
+```wgsl
+// lights.wesl
+module <const MAX: u32 = 8>;
+
+struct Light { pos: vec3f, color: vec3f }
+
+var<private> list: array<Light, MAX>;
+var<private> count: u32;
+
+fn push(l: Light) { list[count] = l; count += 1u; }
+```
+
+- parameters go in a `module` statement near the top of the file. 
+- module params are named, typed, and can have defaults.
+- inside the module, the parameter works as an ordinary `const`. (and is emitted as a WGSL `const`)
+- params with no defaults must be set from host code or via import.
+
+Details
+- multiple parameters can be separated by commas, or multiple `module` statements are allowed, both are equivalent. 
+- `module` statements may not be inside an `@if` condition.
+- `public` marks a parameter as settable from other packages; unmarked parameters are package-internal.
+
+## Conditions
+
+```wgsl
+module <public const SAMPLES: u32 = 8>;
+
+@if(SAMPLES > 4)
+fn blurWide(/* ... */) { /* ... */ }
+```
+
+- `@if`/`@elif` take a translate-time constant expression of type `bool`; identifiers in the expression are module parameters or consts. (condition variables no longer have a separate namespace)
+- A const in an `@if` body can be used in a different condition.
+- A condition or const that transitively depends on itself is an error.
+
+## Importing
+
+```wgsl
+import package::lights<MAX = 32>::{Light, push};
+
+import random::rng<SEED = 1> as r1;
+import random::rng<SEED = 2> as r2;      // two independent instantiations of rng
+import random::rng<SEED = 1+1> as r3;    // alias for r2 (not a third rng)
+```
+
+- arguments are named, not positional. unmentioned parameters take their declared defaults.
+- a plain import (no `< >`) instead takes the link-wide values
+- priority for plain imports: host set default, main module reset default, declaring module default.
+- libraries generally use plain imports so that apps can configure
+- `< >` reads as specialization, like WGSL type parameters
+- const values are const-eval'd
+
+## Module instantiation / concretization when emitting WGSL
+
+When we emit WGSL, we emit a duplicated set of declarations for each set of parameters in use,
+with the exception that types are shared where possible. 
+
+- all `var` declarations are forked per param set
+- types are forked if they depend on a parameter transitively, otherwise shared
+
+Details:
+- implementations can optionally reduce emitted code size by identifying declarations
+  that don't need to be forked (e.g. functions that don't transitively reference
+  a parameter or a forked `var` needn't be duplicated)
+- [ModuleParametersDesign](./ModuleParametersDesign.md) has more examples.
+
+## Module parameters in the main module
+
+- `module` parameters in a main app module are the (only) module params the host can see.
+  (libraries remain encapsulated)
+- The main app module can expose library module parameters to the host
+- The main app module can reset link-wide defaults
+- `private` on an entry sets the value for the link but hides it from the host
+- `@main module` is required for main modules that reset param defaults or expose params from another module to the host. 
+  (And recommended for other main modules. But e.g., a previewer can link and play any module as the main module, so it's okay to link from a main module not tagged `@main`).
+
+```wgsl
+// main.wesl
+@main module <
+  const DEBUG = false,                // param declared locally
+  lygia::tonemap::GAMMA,              // expose a library param to host code
+  bevy_pbr::config::MAX_LIGHTS = 32,  // reset a default
+  private random::rng::SEED = 47,     // reset default, hidden from host
+>;
+```
+
+Host code can set values for main module parameters, e.g.:
+
+```ts
+await link({ ..., params: { DEBUG: true, MAX_LIGHTS: 16 } });
+```
+
+## Laziness and cycles
+
+- imports remain lazy, unused imports are not instantiated.
+- import cycles are still allowed as now, but a referenced import cycle whose arguments don't stabilize around the loop is an error. (a module importing a fixed instantiation of itself is fine)
+- Conditions and consts evaluate on demand, not in separate passes (so that conditions can depend on consts in other conditions)
+
+```wgsl
+// bar.wesl
+module <const C: u32 = 0>;
+
+import package::bar<C = C + 1>::f;   // error if f is used
+```
+
+## Future
+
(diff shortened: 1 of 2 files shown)


k2d222 (approved):
**Form**: the PR doc looks just right: not too verbose, clear enough. In terms of workflow, #222 is basically empty, so I'm wondering if we should just do PR right away for reasonably well-defined proposals. Doesn't matter too much.

**Content**: there are two main differences with #164

* `module <>` syntax instead of `@param`. I don't care much about syntax at this early stage, but my new favorite is a `module { }` block with curly instead of angle brackets. This way, module params are just const declarations inside that block, and we can apply visibility modifiers / re-export mechanisms the same way we do outside. I find it ackward in `< >` that the param const is essentially a declaration, but has no `;`, and it doesn't mix well with the syntax proposed for re-export of submodule params in `@main module <>` IMO.
* params can be overridden in import statements. This was part of #146 (ancestor of #164) but was stripped to simplify things because we were unsure about module instantiation / duplication of variables, and that declaring conditions in shader code + allowing constants in conditions was already a nice improvement. I'm still unsure about the duplication behavior, so I still think we should start without it. For instance, idea of sharing types when possible is interesting. It deserves further discussion as it might cause unexpected things when a type was shared and suddently is no longer unified (e.g., the host code changed a param and that can affect the whole compilation chain).

I don't understand what `@main` does and why we need it. Can you elaborate?


Soo, overall I'm happy with the workflow and approve to merge, and we can continue discussion with the doc in the repo.

stefnotch (approved):
Approved on basis of this proposal being very clear! 

I'll leave my review comments here, commenting on the technical specifics.

proposals/222-ModuleParameters.md
```diff
@@ -0,0 +1,123 @@
...
+fn push(l: Light) { list[count] = l; count += 1u; }
+```
+
+- parameters go in a `module` statement near the top of the file. 
+- module params are named, typed, and can have defaults.
+- inside the module, the parameter works as an ordinary `const`. (and is emitted as a WGSL `const`)
+- params with no defaults must be set from host code or via import.
+
+Details
+- multiple parameters can be separated by commas, or multiple `module` statements are allowed, both are equivalent.
```
stefnotch, line 31: Nitpick: I think having multiple `module` statements would look confusing. It's conceptually just a single module that we have.

proposals/222-ModuleParameters.md
```diff
@@ -0,0 +1,123 @@
...
+
+- parameters go in a `module` statement near the top of the file. 
+- module params are named, typed, and can have defaults.
+- inside the module, the parameter works as an ordinary `const`. (and is emitted as a WGSL `const`)
+- params with no defaults must be set from host code or via import.
+
+Details
+- multiple parameters can be separated by commas, or multiple `module` statements are allowed, both are equivalent. 
+- `module` statements may not be inside an `@if` condition.
+- `public` marks a parameter as settable from other packages; unmarked parameters are package-internal.
```
stefnotch, line 33: Nitpick: Should we really tie visibility into the module parameter design? It makes the feature more complex. 
What does it bring us?

proposals/222-ModuleParameters.md
```diff
@@ -0,0 +1,123 @@
...
+Host code can set values for main module parameters, e.g.:
+
+```ts
+await link({ ..., params: { DEBUG: true, MAX_LIGHTS: 16 } });
+```
+
+## Laziness and cycles
+
+- imports remain lazy, unused imports are not instantiated.
+- import cycles are still allowed as now, but a referenced import cycle whose arguments don't stabilize around the loop is an error. (a module importing a fixed instantiation of itself is fine)
```
stefnotch, line 108: Nitpick: "don't stabilize" is slightly challenging to implement. It implies being able to solve the Collatz conjecture.

stefnotch: I thought this was a very funny comment, so I felt inspired to actually demonstrate it

```wgsl
// We do two wildcard imports. 
// If they refer to the same thing it's fine.
// But we have to check if they really refer to the same thing.
import package::bar::*;
import package::collatz::*;

// Usage of A
const B = A + 1;



// bar.wesl
// So this is the trivial setup
public const A = 3;



// collatz.wesl
// And here we have a parameterized module
module <const N: u32>;

// Base case. We re-export the A from bar.wesl
@if(N == 1)
public import package::bar::A; // this one is okay

// Use an arbitrary number here
@if(N == 24524623)
public const A = 7; // this one clashes

// And drive the collatz conjecture logic
// This repeatedly imports a different collatz.wesl
// And the collatz conjecture says that it should stabilize at 1
public import package::collatz<select(N/2, 3*N+1, N % 2 == 1)>::A;
```

proposals/222-ModuleParameters.md
```diff
@@ -0,0 +1,123 @@
...
+- Conditions and consts evaluate on demand, not in separate passes (so that conditions can depend on consts in other conditions)
```
stefnotch, line 109: Nitpick: This is a bit unclear.

```
const A = 3;

// this should be disallowed
@if(A)
const B = 5;

// but that should be allowed?
import package::foo<A>;
```

proposals/222-ModuleParameters.md
```diff
@@ -0,0 +1,123 @@
...
+- libraries generally use plain imports so that apps can configure
+- `< >` reads as specialization, like WGSL type parameters
+- const values are const-eval'd
+
+## Module instantiation / concretization when emitting WGSL
+
+When we emit WGSL, we emit a duplicated set of declarations for each set of parameters in use,
+with the exception that types are shared where possible. 
+
+- all `var` declarations are forked per param set
```
stefnotch, line 70: This is the one thing that worries me. We're deviating from well studied module systems with this one rule.

WGSL currently relies on shared state quite often. Like `var<private>` or [like this piece of code that does a bit of memory management on the GPU](https://gitlab.tuwien.ac.at/e193-02-visvu-students/visvu-2025-trautner_bruckner_2021-brandmair-elagabani/-/blob/main/src/shaders/tiling.wgsl?ref_type=heads#L167). I don't think I could realistically reason about how such code would behave if the `var` declarations were shared across multiple instances. 

Especially since the sharing would be a global property, as opposed to a local one. Two completely unrelated places can import a `my_buffer<u32>` module and would see the data from the other place. 

We should enable local reasoning. There are two ways of getting there, IMO.

1. Discourage module level `var`s. This is the world where binding structs, #200 and co live. Explicitly pass the `var`s around.
2. Learn from languages that do in fact have a well thought out module system. OCaml distinguishes between "this is a module" and "that creates a module". OCaml goes many steps further, each of which enables completely new and lovely ways of composing code. We should learn from there!

proposals/222-ModuleParameters.md
```diff
@@ -0,0 +1,123 @@
...
+  that don't need to be forked (e.g. functions that don't transitively reference
+  a parameter or a forked `var` needn't be duplicated)
+- [ModuleParametersDesign](./222-ModuleParametersDesign.md) has more examples.
+
+## Module parameters in the main module
+
+- `module` parameters in a main app module are the (only) module params the host can see.
+  (libraries remain encapsulated)
+- The main app module can expose library module parameters to the host
+- The main app module can reset link-wide defaults
```
stefnotch, line 84: Nitpick: "Link wide defaults" is messing with me. 

As a quick, not too hypothetical example:
I write a PBR shader and rely on Lygia's gamma correction. I also write a texture fetching routine and rely on Lygia's gamma correction. I expect the gamma value to be something very specific, since the PBR shader's math is based on that. And I expect the gamma value to be something very specific in the texture fetching shader, since that gamma value is specified in some external standard.

So I now have to fully specify the gamma value and cannot rely on the default. If I accidentally rely on the default, that is a bug waiting to happen. I have to remember to check each module for "wait, does this have behaviour that anyone else could override". I think an explicit world is simpler and more convenient to use.

proposals/222-ModuleParameters.md
```diff
@@ -0,0 +1,123 @@
...
+- `< >` reads as specialization, like WGSL type parameters
+- const values are const-eval'd
+
+## Module instantiation / concretization when emitting WGSL
+
+When we emit WGSL, we emit a duplicated set of declarations for each set of parameters in use,
+with the exception that types are shared where possible. 
+
+- all `var` declarations are forked per param set
+- types are forked if they depend on a parameter transitively, otherwise shared
```
stefnotch, line 71: Mostly unimportant design question: I think this affects name mangling, which affects our shared test suite. I guess we'll have to go down the road of having a shared test suite that is name mangling agnostic?

proposals/222-ModuleParametersDesign.md
```diff
@@ -0,0 +1,200 @@
...
+
+### Authors can share more or less if they want
+
+With those rules for when to share vars, there's still room for author control:
+
+- State is unique per module parameterization, but if an author wants to share state across concrete modules,
+  they can do so by hoisting the var into another module w/o the module parameter.
+- State is shared per module-parameterization set, but if an author wants unique state per importer, 
+  they have two options: pass in the state via a pointer; or add an additional module parameter
+  `rng<STREAM = 47>` vs `rng<STREAM = 19>`.
```
stefnotch, line 150: Here's a good example for where "sharing per parameterization" breaks down.

When two unrelated modules happen to import a module with the same parameterization
```
// perlin_noise.wesl
import math::rng<STREAM = 42>

fn get_value(a: vec2f) -> f32 { ... }
```

```
// simplex_noise.wesl
import math::rng<STREAM = 42>

fn get_value(a: vec2f) -> f32 { ... }
```

Then we get the insane behaviour where calling
```
perlin_noise::get_value()
simplex_noise::get_value()
```

gives different results from calling
```
simplex_noise::get_value()
perlin_noise::get_value()
```

proposals/222-ModuleParametersDesign.md
```diff
@@ -0,0 +1,200 @@
...
+
+## Module parameters designs in other languages
```
stefnotch, line 152: And interestingly Rust also wants module level generics

https://smallcultfollowing.com/babysteps/blog/2022/04/17/coherence-and-crate-level-where-clauses/#module-level-generics

